Issue #6, opened 2020-05-06
Reduce logging volume


jkemp101:
Large Sync Engine clusters can produce an enormous volume of logs. https://github.com/closeio/nylas/pull/36 reduced this a bit but more work is needed to further reduce logging frequency. This may make it more difficult to troubleshoot some syncing issues so we could add the ability to enable per account debug logging.

dggreenbaum:
This is significantly less important after moving to Loki. Loki ingests logs for ~15% of the cost of doing it in ES for the same log volume.

PR #17, opened 2020-07-30
Log polling statement every 5 minutes to cut down on logging volume

jkemp101:
This might be a quick win to reduce logging but still get an idea if an account is actively polling.

Ref https://github.com/closeio/sync-engine/issues/6

### DIFF
--- a/inbox/mailsync/backends/imap/generic.py
+++ b/inbox/mailsync/backends/imap/generic.py
@@ -347,8 +347,9 @@ class FolderSyncEngine(Greenlet):
         log.bind(state='poll')
         # Only log every 5 minutes to cut down on the volume of
         # this log statement
-        if time.time() - self.polling_logged_at > 60 * 5:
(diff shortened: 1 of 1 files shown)


inbox/mailsync/backends/imap/generic.py
```diff
@@ -343,7 +345,11 @@ def initial_sync(self):
     @retry_crispin
     def poll(self):
         log.bind(state='poll')
-        log.debug('polling')
+        # Only log every 5 minutes to cut down on the volume of
+        # this log statement
+        if time.time() - self.polling_logged_at > 60 * 5:
```
thomasst, line 350: Minor: Could save a call by doing `timestamp = time.time()` and using it in both places.

jkemp101: I realized that after looking at the PR, you guilted me into fixing it.

PR #20, opened 2020-08-04
Improve `fetch_corresponding_thread` optimization

thomasst:
This iterates on the [previous commit ](https://github.com/closeio/sync-engine/commit/55298ee30beb6145e1e880e7b486452934e9084d) by optimizing & limiting the database
query to avoid memory issues.

### DIFF
--- a/inbox/util/threading.py
+++ b/inbox/util/threading.py
@@ -15,6 +15,13 @@ MAX_MESSAGES_SCANNED = 20000
 def fetch_corresponding_thread(db_session, namespace_id, message):
     """fetch a thread matching the corresponding message. Returns None if
        there's no matching thread."""
+    # handle the case where someone is self-sending an email.
+    if not message.from_addr or not message.to_addr:
+        return None
+
+    message_from = [t[1] for t in message.from_addr]
+    message_to = [t[1] for t in message.to_addr]
+
     # FIXME: for performance reasons, we make the assumption that a reply
     # to a message always has a similar subject. This is only
     # right 95% of the time.
@@ -58,14 +65,8 @@ def fetch_corresponding_thread(db_session, namespace_id, message):
                 else:
                     return match.thread
 
-            # handle the case where someone is self-sending an email.
-            if not message.from_addr or not message.to_addr:
-                return
-
             match_from = [t[1] for t in match.from_addr]
             match_to = [t[1] for t in match.from_addr]
-            message_from = [t[1] for t in message.from_addr]
-            message_to = [t[1] for t in message.to_addr]
 
             if (len(message_to) == 1 and message_from == message_to and
                     match_from == match_to and message_to == match_from):
@@ -77,4 +78,4 @@ def fetch_corresponding_thread(db_session, namespace_id, message):
                 else:
                     return match.thread
 
-    return
+    return None
--- a/inbox/util/threading.py
+++ b/inbox/util/threading.py
@@ -27,10 +27,11 @@ def fetch_corresponding_thread(db_session, namespace_id, message):
     # right 95% of the time.
     clean_subject = cleanup_subject(message.subject)
 
-    # XXX: It is much faster to sort client-side by message date, we therefore
-    # use `contains_eager` and `outerjoin` instead of `joinedload` to fetch
-    # the messages by thread in no particular order. We also use a limit to
-    # avoid scanning too many / large threads.
+    # XXX: It is much faster to sort client-side by message date. We therefore
+    # use `contains_eager` and `outerjoin` to fetch the messages by thread in
+    # no particular order (as opposed to `joinedload`, which would use the
+    # order_by on the Message._thread backref).  We also use a limit to avoid
+    # scanning too many / large threads.
     threads = db_session.query(Thread). \
         filter(Thread.namespace_id == namespace_id,
                Thread._cleaned_subject == clean_subject). \
--- a/inbox/util/threading.py
+++ b/inbox/util/threading.py
@@ -20,7 +20,7 @@ def fetch_corresponding_thread(db_session, namespace_id, message):
     # right 95% of the time.
     clean_subject = cleanup_subject(message.subject)
 
-    # XXX: It is much faster to sort client-side by message ID, we therefore
+    # XXX: It is much faster to sort client-side by message date, we therefore
     # use `contains_eager` and `outerjoin` instead of `joinedload` to fetch
     # the messages by thread in no particular order. We also use a limit to
     # avoid scanning too many / large threads.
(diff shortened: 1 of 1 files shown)


inbox/util/threading.py
```diff
@@ -16,25 +19,24 @@ def fetch_corresponding_thread(db_session, namespace_id, message):
...
     clean_subject = cleanup_subject(message.subject)
+
+    # XXX: It is much faster to sort client-side by message ID, we therefore
+    # use `contains_eager` and `outerjoin` instead of `joinedload` to fetch
+    # the messages by thread in no particular order. We also use a limit to
+    # avoid scanning too many / large threads.
     threads = db_session.query(Thread). \
         filter(Thread.namespace_id == namespace_id,
                Thread._cleaned_subject == clean_subject). \
+        outerjoin(Message, Thread.messages). \
```
wojcikstefan, line 30: Do we really need an *outer* join here? We don't really have a need for threads that contain no messages, do we?

jkemp101: Same question. If this helps with the query plan we should just document that so we remember in the future.

inbox/util/threading.py
```diff
@@ -16,25 +19,24 @@ def fetch_corresponding_thread(db_session, namespace_id, message):
...
-        for match in thread.messages:
-            # If we've scanned more than `MAX_MESSAGES_SCANNED`, give up and
-            # assume that a matching thread doesn't exist. We do this because
-            # the number of threads and messages we iterate over can get really
-            # out of hand for some transactional emails.
-            num_messages_scanned += 1
-            if num_messages_scanned > MAX_MESSAGES_SCANNED:
-                return
-
+        messages = sorted(thread.messages, key=attrgetter('received_date'))
```
jkemp101, line 38: I feel like I'm missing something. Where was this ordering being done before? I don't see it in the `order_by` above.

inbox/util/threading.py
```diff
@@ -16,25 +19,24 @@ def fetch_corresponding_thread(db_session, namespace_id, message):
     # to a message always has a similar subject. This is only
     # right 95% of the time.
     clean_subject = cleanup_subject(message.subject)
+
+    # XXX: It is much faster to sort client-side by message ID, we therefore
```
jkemp101, line 23: Another option would be the break this apart into two queries and first get a list of thread ids and then for each thread query the messages. This would let us apply limits to both and let ordering be done in the DB. Using a `limit` and then sorting client side made me think about this since that approach could end up with different results since the `limit` is applied before the sorting.

inbox/util/threading.py
thomasst: It matches the previous query generated by SQLAlchemy, and in general how loads work (if you fetch threads with messages you'd expect to have all the threads, even if they have no messages). If you specifically wanted to include threads without messages, you could use `join`. But since all threads have messages anyways, why change the type of join?

inbox/util/threading.py
thomasst: Yes. It used `joinedload` before, which takes the sort from the relation (https://github.com/closeio/sync-engine/blob/master/inbox/models/message.py#L80)

inbox/util/threading.py
thomasst: > that approach could end up with different results since the `limit` is applied before the sorting

Only within threads. We could skip the last thread if you're concerned.

thomasst:
```
>>> print(db_session.query(Thread). \
...         filter(Thread.namespace_id == namespace_id,
...                Thread._cleaned_subject == clean_subject). \
...         outerjoin(Message, Thread.messages). \
...         order_by(desc(Thread.id)). \
...         options(load_only('id', 'discriminator'),
...                 contains_eager(Thread.messages).load_only(
...                     'from_addr', 'to_addr', 'bcc_addr', 'cc_addr', 'received_date')). \
...         limit(20000))
SELECT thread.type AS thread_type, thread.id AS thread_id, message.id AS message_id, message.from_addr AS message_from_addr, message.to_addr AS message_to_addr, message.cc_addr AS message_cc_addr, message.bcc_addr AS message_bcc_addr, message.received_date AS message_received_date 
FROM thread LEFT OUTER JOIN message ON message.thread_id = thread.id 
WHERE thread.namespace_id = :namespace_id_1 AND thread._cleaned_subject = :_cleaned_subject_1 ORDER BY thread.id DESC
 LIMIT :param_1
>>> print(db_session.query(Thread). \
...         filter(Thread.namespace_id == namespace_id,
...                Thread._cleaned_subject == clean_subject). \
...         order_by(desc(Thread.id)). \
...         options(load_only('id', 'discriminator'),
...                 joinedload(Thread.messages).load_only(
...                     'from_addr', 'to_addr', 'bcc_addr', 'cc_addr')))
SELECT thread.type AS thread_type, thread.id AS thread_id, message_1.id AS message_1_id, message_1.from_addr AS message_1_from_addr, message_1.to_addr AS message_1_to_addr, message_1.cc_addr AS message_1_cc_addr, message_1.bcc_addr AS message_1_bcc_addr 
FROM thread LEFT OUTER JOIN message AS message_1 ON message_1.thread_id = thread.id 
WHERE thread.namespace_id = :namespace_id_1 AND thread._cleaned_subject = :_cleaned_subject_1 ORDER BY thread.id DESC, message_1.received_date
>>> 
```

inbox/util/threading.py
wojcikstefan: > XXX: It is much faster to sort client-side by message ID

What do you mean by "by message ID"? We order by message's `received_date` below.

inbox/util/threading.py
```diff
@@ -16,25 +19,24 @@ def fetch_corresponding_thread(db_session, namespace_id, message):
...
+    # use `contains_eager` and `outerjoin` instead of `joinedload` to fetch
+    # the messages by thread in no particular order. We also use a limit to
```
wojcikstefan, line 25: I think it's worth mentioning that with `joinedload`, the order would come from the `_thread` relation's backref's `order_by`.

inbox/util/threading.py
thomasst: corrected in d0b1d1239a2096bdbe95efe3c96ad893149a34fc

inbox/util/threading.py
```diff
@@ -12,29 +15,36 @@
...
     # FIXME: for performance reasons, we make the assumption that a reply
     # to a message always has a similar subject. This is only
     # right 95% of the time.
     clean_subject = cleanup_subject(message.subject)
+
+    # XXX: It is much faster to sort client-side by message date. We therefore
+    # use `contains_eager` and `outerjoin` to fetch the messages by thread in
+    # no particular order (as opposed to `joinedload`, which would use the
+    # order_by on the Message._thread backref).  We also use a limit to avoid
+    # scanning too many / large threads.
```
wojcikstefan, line 34: Now that's a great comment 👌

PR #21, opened 2020-08-04
Explicit `return None` + optimize the case of self-sending

wojcikstefan:
This started off as a PR removing an unnecessary `return` at the end of the function and then turned into something else entirely (see the commit messages for the play-by-play history).

@alecrosenbaum said:
> Personally I prefer an explicit `return None` because it's less likely that
> I'll miss it when reading through a function.

I cannot personally relate to it, but if there's anyone who benefits from `return None`, then I'm happy to oblige.

While making this change, I also noticed that one of the `return` statements within the thread-match loop only looked at the `message` param and didn't need any additional data (also, none of the `return` statements above it could've matched if the `message` didn't have a `from_addr` and a `to_addr`). Because of that, we can move that check all the way up and hence, in case of self-sent emails, we don't even need to fetch anything from the database.

### DIFF
--- a/inbox/util/threading.py
+++ b/inbox/util/threading.py
@@ -18,6 +18,12 @@ def fetch_corresponding_thread(db_session, namespace_id, message):
     # FIXME: for performance reasons, we make the assumption that a reply
     # to a message always has a similar subject. This is only
     # right 95% of the time.
+
+    # Handle the case where someone is self-sending an email. None of the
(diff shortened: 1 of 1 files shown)


thomasst:
Since this will conflict with https://github.com/closeio/sync-engine/pull/20, I added it there (https://github.com/closeio/sync-engine/pull/20/commits/d39004b7eb516b5298181da7d0003d21dc502711).

PR #22, opened 2020-08-13
Facelift

AlecRosenbaum:
This change:
* Pulls missing tests from `tests/` into `inbox/test`
* removes the top level `tests/` folder
* removes `requirements.txt` in favor of `requirements_frozen.txt`
* removes unused php-based testing and linting scripts
* removes `.lgtm` config
* removes debian package publishing information
* removes `MANIFEST.in`, which I'm guessing was related to debian releases?
* changes the test running steps to use a docker image (not pushed anywhere), which makes local testing consistent and much easier
  - The db setup function and default test config was slightly changed to accommodate this
  - I had to skip one test which started failing with the new docker image for some reason. I think it
    has something to do with how mysql is set up, but I'm not sure. I'll have to investigate later when I look through some of the other skipped tests.


AlecRosenbaum:
> > Pulls missing tests from tests/ into inbox/test
> I trust you did that right :)

I hope so :) I basically just did a `git diff --no-index tests inbox/test`. Most changes were to imports, but there were also some tests where the version in `inbox/test` seemed newer (which I left). There were only a few test cases that were completely missing from `inbox/test` and needed to be added.

> Mayyybe is this related at all? https://github.com/closeio/nylas/pull/54/files

Yeah, it definitely could be. The versions of a bunch of these packages should probably be bumped at some point soon, so we'll have to deal with it then (if not sooner).

Issue #12, opened 2020-07-09
Update linters


jkemp101:
Add `static-code-analysis` circle job that includes our standard linters

- [ ] flake8
- [x] black
- [x] isort

jkemp101:
iirc flake8 identified so many issues it wasn't worth enabling yet

squeaky-pl:
Superseded by https://github.com/closeio/sync-engine/issues/311

PR #23, opened 2020-08-17
Format the codebase

AlecRosenbaum:
Related: https://github.com/closeio/sync-engine/issues/12

This change:
* adds a `requirements_lint.txt` file for linter requirements
* formats code with `black`
* adds a `static-code-analysis` job to CircleCI, which enforces code formatting

This is a pretty big diff which murders git history, but the files look much prettier now IMO. Note that the formatting is done with the default `black` settings. This means there's a line-length of `88` and strings are normalized.


jkemp101:
Do we want to switch to double quotes for this repo?

AlecRosenbaum:
I think so. Trying to adhere to a formatting convention explicitly not supported by `black` is not great IMO. In other projects where quotes aren't enforced due to the single quote convention, I have spent what felt like completely wasted time manually going back through changes and swapping quotes.

PR #25, opened 2020-08-19
Move the tests folder

AlecRosenbaum:
This change moves the tests folder from `inbox/test` to `tests/`.

This is consistent with other `closeio` projects, and offers a few benefits. Primarily being that it's easy to exclude test files from code searches.


thomasst (approved):
LGTM, and the tests pass!

PR #29, opened 2020-09-22
Always require an encrypted connection for IMAP & SMTP

thomasst:
Requires https://github.com/closeio/sync-engine/pull/28 first (because the migrations are in sequence).

Deploy PR first, and only run migration when it's fully rolled out.

### DIFF
--- a/inbox/auth/base.py
+++ b/inbox/auth/base.py
@@ -1,5 +1,7 @@
 from inbox.basicauth import NotSupportedError
 
+from .utils import create_imap_connection
+
 
 def handler_from_provider(provider_name):
     """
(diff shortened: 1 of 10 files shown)


AlecRosenbaum (approved):
Have you searched our logs to double check that we haven't seen any of the `"STARTTLS supported but failed for SSL NOT required authentication"` messages?

thomasst:
> Have you searched our logs to double check that we haven't seen any of the `"STARTTLS supported but failed for SSL NOT required authentication"` messages?

I haven't seen it. I also don't think we have any non-SSL accounts.

PR #30, opened 2020-09-22
Outlook support

thomasst:
Tested by successfully authorizing an Outlook account using `./bin/inbox-auth` & verifying it syncs.


inbox/auth/oauth.py
```diff
@@ -90,29 +91,24 @@ def authenticate_imap_connection(self, account, conn):
...
+            log.error("user_info_fetch_failed", error_code=e.code, error=e)
+            raise ConnectionError()
+        except urllib.error.URLError as e:
             log.error("user_info_fetch_failed", error=e)
             raise ConnectionError()
 
-        userinfo_dict = response.json()
-
-        if "error" in userinfo_dict:
-            assert userinfo_dict["error"] == "invalid_token"
```
AlecRosenbaum, line 105: That's an oddly specific condition, any idea what this was testing against?

inbox/models/backends/gmail.py
```diff
@@ -25,33 +25,18 @@
...
+    OAUTH_CLIENT_SECRET = config.get_required("GOOGLE_OAUTH_CLIENT_SECRET")
+
     id = Column(ForeignKey(ImapAccount.id, ondelete="CASCADE"), primary_key=True)
 
     __mapper_args__ = {"polymorphic_identity": "gmailaccount"}
 
     client_id = Column(String(256))
     scope = Column(String(512))
 
-    # XXX: These fields are not currently used.
```
AlecRosenbaum, line 41: weren't these already dropped? https://github.com/closeio/sync-engine/pull/28

thomasst: I just changed base for this PR. Reload please.

inbox/models/backends/outlook.py
```diff
@@ -1,21 +1,25 @@
...
 
     __mapper_args__ = {"polymorphic_identity": "outlookaccount"}
 
-    # STOPSHIP(emfree) store these either as secrets or as properties of the
-    # developer app.
     client_id = Column(String(256))
-    client_secret = Column(String(256))
     scope = Column(String(512))
+
+    # TODO: These fields are unused.
```
AlecRosenbaum, line 21: dropping these in a future change?

AlecRosenbaum (approved):

inbox/auth/oauth.py
thomasst: - The expected OpenID response would be a 401, like this: https://openid.net/specs/openid-connect-core-1_0.html#UserInfoError
- I found this doc in their YouTube API that suggests it returns a 400 with `invalid_token`. https://developers.google.com/youtube/v3/live/guides/auth/client-side-web-apps 

I can do a test to see what Google actually returns.

It also seems that the original code doesn't check the HTTP response code, which is why it looks the error response. I don't think the original code is particularly great.

inbox/models/backends/outlook.py
thomasst: Yep.

inbox/auth/oauth.py
AlecRosenbaum: Yep agreed that the change is better, I was mostly just curious.

thomasst:
closing in favor of https://github.com/closeio/sync-engine/pull/31

PR #38, opened 2020-10-08
Truncate long UIDs

thomasst:
Fixes "Data too long for column 'uid' at row 1" issue for an account we're syncing that has a UID 1034 bytes in length.


jkemp101:
No concerns this could cause duplicates?

thomasst:
Right now we only have one case where the UID is too long. I'm not too concerned about it and don't think we need to handle events with long UIDs perfectly.

jkemp101 (approved):

PR #39, opened 2020-10-08
Update MySQL to 5.7.31

dggreenbaum:
Updates MySQL to the latest 5.7 build at time of writing.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       - 6379
 
   mysql:
-    image: mysql:5.7.19
+    image: mysql:5.7.31
     environment:
       MYSQL_ROOT_PASSWORD: nylas_it
       MYSQL_USER: inboxtest


jkemp101:
I did a review of the major changes for each release and nothing jumped out. I think we went from 5.6 to 5.7 without issues so hopefully this would be uneventful also.

jkemp101 (approved):

thomasst (approved):
We'd want to change it here as well:

https://github.com/closeio/devtools/blob/master/docker-compose.yaml#L396

I haven't encountered any issues (and wouldn't expect any).

I did get this warning repeatedly after running the updated MySQL version:

```
nylas_mysql_1                 | 2020-10-08T21:29:29.178595Z 0 [Warning] InnoDB: Table mysql/innodb_index_stats has length mismatch in the column name table_name.  Please run mysql_upgrade
```

dggreenbaum:
@thomasst it looks like RDS will run the `mysql_upgrade` command during the upgrade.

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.MySQL.html

I'll test this against a snapshot of a prod instance to make sure we don't get these warnings.

dggreenbaum:
devtools PR: https://github.com/closeio/devtools/pull/158

PR #45, opened 2020-12-15
Allow users to set nylas SMTP settings

zkryakgul:
Summary:
This changes allows users to set nylas's generic account SMTP settings without change the code. And also allows you(close.io settings) to use it with default settings if you don't send the smtp data.

### DIFF
--- a/inbox/api/srv.py
+++ b/inbox/api/srv.py
@@ -152,10 +152,10 @@ def ns_all():
 def _get_account_data_for_generic_account(data):
     email_address = data["email_address"]
     sync_email = data.get("sync_email", True)
-    smtp_server_host = data["smtp_server_host"] or "localhost"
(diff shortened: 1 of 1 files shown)


inbox/api/srv.py
```diff
@@ -152,16 +152,21 @@ def ns_all():
 def _get_account_data_for_generic_account(data):
     email_address = data["email_address"]
     sync_email = data.get("sync_email", True)
+    smtp_server_host = data["smtp_server_host"] or "localhost"
```
jkemp101, line 155: Did you run tests for this change and potentially add/modify them to cover this use case? I think these should be `data.get()`'s like the line above.

zkryakgul: I updated the lines and run the tests. They passed with no error. Sorry for the mistake. I was thought you just sending blank the SMTP values with keys.

PR #49, opened 2021-07-02
Cleanup Readme

jkemp101:


dprkr:
Hi I see vagrant has been mostly removed from the readme in favour of docker.  Can you provide some guidance on docker?  I built the image and then tried `docker run -d -p 5555:5555 {imageid}` but the container just exits immediately with no logs.  I'm new to docker. Thx

jkemp101:
@dprkr Unfortunately we don't have a full set of updated instructions to get you started with Docker. So you are going to have to figure out some of this by trial and error. 

This command is probably closer to what you are looking for but it is just going to fail with an error because the config files in the image aren't going to be setup for your environment. Theoretically if you got everything configured properly and build an image with your settings, you could start one of the processes like this:
```
docker run -d -p 5555:5555 {imageid} bash -c "PYTHONPATH=. /src/bin/inbox-start"
``` 

What you probably want to do is to start a container like this so you can poke around the various `bin/*` scripts. But of course any file changes (e.g. configs) will be loss when the container is recreated. 
```
docker run -d -p 5555:5555 -t -i {imageid}
```

dprkr:
Thanks for this

PR #52, opened 2021-09-09
reset mimepart.body

froxCZ:


inbox/models/message.py
```diff
@@ -340,6 +340,21 @@ def create_from_synced(cls, account, mid, folder_name, received_date, body_strin
...
+                    # each having their own copy of the actual data they refer too.
+                    # Here is the part where Mimepart.stream keeps the reference
+                    # to copy of body, and never removes it later. If the body is
+                    # 16MB of each mimepart, and we have thousands of mimeparts,
+                    # we run out of memory.
+                    # https://github.com/closeio/flanker/blob/master/flanker/mime/message/part.py#L75
+                    # By resetting the body attribute we remove the data copy for that mimepart.
+                    # In case the prop was accessed after that, it would just be
+                    # loaded again. So we trade cpu for memory
+                    mimepart.body = None
```
froxCZ, line 357: Ideally flanker would not keep a list, but just a stream of mimeparts, and thus remove them once they are processed in the loop. But that would be a bigger change imo and in flanker

thomasst: Should this be `mimepart._body = None` so we don't overwrite the property itself?

thomasst: This looks good, but maybe we should just say `mimepart._body = None` so we don't set `_body_changed` (since we don't actually change the body), just in case.

thomasst (approved):

inbox/models/message.py
froxCZ: That makes sense. `    mimepart._container._body = None`

froxCZ:
This did not help, our code buffers all the mimepart bodies anyway. More at https://github.com/closeio/sync-engine/pull/53

PR #53, opened 2021-09-10
Parse only 10 mimeparts

froxCZ:
Tested locally, email was still received. Will merge after tested in prod.

### DIFF
--- a/inbox/models/message.py
+++ b/inbox/models/message.py
@@ -317,7 +317,15 @@ class Message(MailSyncBase, HasRevisions, HasPublicID, UpdatedAtMixin, DeletedAt
         if parsed is not None:
             plain_parts = []
             html_parts = []
-            for mimepart in parsed.walk(with_self=parsed.content_type.is_singlepart()):
+            # Parse only 10 mimeparts, because flanker is very inefficient
+            # in parsing and can get OOM.
+            # We probably don't use body snippet and body anyway, so we plan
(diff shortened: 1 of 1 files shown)


thomasst (approved):

inbox/models/message.py
```diff
@@ -317,7 +317,15 @@ def create_from_synced(cls, account, mid, folder_name, received_date, body_strin
...
             html_parts = []
-            for mimepart in parsed.walk(with_self=parsed.content_type.is_singlepart()):
+            # Parse only 10 mimeparts, because flanker is very inefficient
+            # in parsing and can get OOM.
+            # We probably don't use body snippet and body anyway, so we plan
+            # to remove this loop completely.
+            for idx, mimepart in enumerate(
+                parsed.walk(with_self=parsed.content_type.is_singlepart())
+            ):
+                if idx >= 10:
```
jkemp101, line 327: @thomasst Should we create an issue in this repo for this to further improve it? Would we want sync-engine to still create snippets but have an option to turn that functionality off?

thomasst: - We should split out higher level email parsing (that we use in Close) into a separate library. Flanker has its issues and I think we shouldn't use it.
- I don't see much value in the snippet. In general it depends on what use cases we want to support with sync-engine. For example, we don't need to store or search attachments and bodies, but not every user of sync-engine might want to re-parse emails.

PR #54, opened 2021-09-17
New Dockerfile

squeaky-pl:


Dockerfile
```diff
@@ -1,6 +1,7 @@
-FROM ubuntu:16.04
+FROM ubuntu:xenial-20210804
```
squeaky-pl, line 1: It felt appropriate to pin this, as this is new base image

.dockerignore
```diff
@@ -0,0 +1,6 @@
+.circleci
```
squeaky-pl, line 1: Added this so the layer caching works nicely

Dockerfile
```diff
@@ -30,11 +32,31 @@ RUN DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get dist-upgrade -y
...
+WORKDIR /opt/app
+
+ENV PATH="/opt/venv/bin:$PATH"
+
 COPY ./ ./
-RUN pip install setuptools==44.0.0
-RUN pip install pip==20.3.4
-RUN pip install -r requirements_frozen.txt -e .
+RUN \
+  virtualenv /opt/venv && \
```
AlecRosenbaum, line 56: why use a virtualenv inside a docker image?

squeaky-pl: This was a suggestion from @jkemp101

docker-compose.yml
```diff
@@ -15,12 +15,13 @@ services:
...
 
-  nylas:
+  app:
     build:
       context: .
       dockerfile: Dockerfile
     volumes:
-      - .:/src
+      - .:/opt/app
+    tmpfs: /opt/app/inbox_sync.egg-info
```
squeaky-pl, line 24: I had to do this so `pip install -e .` keeps working after mounting the source code volume now that we use non-root user and Docker on Linux mounts volumes as root: https://github.com/moby/moby/issues/2259

Another alternative would be to chown to 5000:5000 the code on the host but I found that ugly.

.circleci/config.yml
```diff
@@ -54,6 +68,4 @@ workflows:
       - static-code-analysis:
           context:
             - docker-hub-creds-ro
-      - build:
-          context:
```
squeaky-pl, line 58: Had to remove this because it shadows the variables added on project level.

.circleci/config.yml
```diff
@@ -28,24 +28,38 @@ jobs:
...
           name: run tests
           command: |
             docker login -u $DOCKER_USER -p $DOCKER_PASS
-            docker-compose run nylas bash -ec '
+            docker-compose run app bash -ec '
               pip install -e . \
               && bin/wait-for-it.sh mysql:3306 \
               && NYLAS_ENV=test py.test tests/
             '
+      - run:
```
squeaky-pl, line 49: Copied this from https://github.com/closeio/authalligator/blob/master/.circleci/config.yml#L98-L113

.circleci/config.yml
AlecRosenbaum: should we remove it for `static-code-analysis` too?

.circleci/config.yml
```diff
@@ -28,24 +28,38 @@ jobs:
...
+          name: Push the tested image to docker hub
+          environment:
+            PROJECT: sync-engine
+            PROJECT_NAMESPACE: closeio
+          command: |
+            PROJECT_VSN="$CIRCLE_BRANCH"
+            PROJECT_SHA="$CIRCLE_SHA1"
+            docker login -u $DOCKER_USER -p $DOCKER_PASS
+            # tag and push both sync-engine:master and sync-engine:<sha1>
+            docker tag "${PROJECT}_app" "${PROJECT_NAMESPACE}/${PROJECT}:branch-${PROJECT_VSN}"
```
AlecRosenbaum, line 59: where are `PROJECT` and `PROJECT_NAMESPACE` set?

squeaky-pl: @AlecRosenbaum not sure if I understand correctly but it's set under `enviornment:` couple of lines above.

AlecRosenbaum: 🤦‍♂️ Yep, my bad.

AlecRosenbaum (approved):

Dockerfile
```diff
@@ -1,6 +1,7 @@
-FROM ubuntu:16.04
+FROM ubuntu:xenial-20210804
 
-USER root
+RUN groupadd -g 5000 sync-engine \
+  && useradd -d /home/sync-engine -m -u 5000 -g 5000 sync-engine
```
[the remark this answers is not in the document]
squeaky-pl: @jkemp101 Is this a concern really? I'd rather not introduce another symbol. A typical Debian install has NAME_REGEX="^[a-z][-a-z0-9_]*\$" and dashes were allowed for example `www-data` is commonly used for Nginx and Apache.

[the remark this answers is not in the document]
squeaky-pl: This specific Dockerfile won't run on any *NIX anyway because it contains references to Debian specific package names.

jkemp101 (approved):

PR #55, opened 2021-09-23
Provide a possibility of running arbitrary startup hook 

squeaky-pl:
A third party can set `SYNC_ENGINE_STARTUP_HOOK` environment variable to `module.submodule.callable`. This will be executed once per process as part of the startup and can serve as a way to customize the process behavior.


squeaky-pl:
After discussions it was decided not to go this route

PR #58, opened 2021-09-27
Drop tox

squeaky-pl:
Tox is a convenient way to run arbitrary commands in isolated environments to serve as a single entry point for build, test and release activities.

But we are not actively using/maintaining support for it and the tox file included in this repository simply does not work with current Docker setup. It would be better to drop it all together.


requirements_frozen.txt
```diff
@@ -108,10 +108,8 @@ git+https://github.com/mhahnenberg/sqlalchemy.git@e5f7586346ce5fafe2c182f70bf6ef
 statsd==3.1
 structlog==17.2.0
 tldextract==1.7.5
-tox==2.3.1
 uritemplate==0.6
 urllib3==1.22
-virtualenv==15.1.0
```
squeaky-pl, line 114: This is only a dependency of tox. Tox creates an isolated virtualenv to run each command.

AlecRosenbaum (approved):

PR #59, opened 2021-09-27
Bring back Flake8 

squeaky-pl:
Some parts of the code hint me that there used to be Flake8 running on this code. For the sake of more safety let's bring that static analysis back. I brought back the latest version of Flake8

Problems that needed to be fixed:
- comparing `None` and `True` with `==` in sqlalchemy queries
- dead imports
- mixing tabs and spaces, missing new lines
- incorrect string escapes in regexps such as `"\s"` (needs to be either `"\\s"` or `r"\s"`)
- nondescriptive var names

One particularly annoying problem I had to solve is incorrectly using imports to make pytest fixtures visible instead of relying on coftests.py / plugins. This chokes flake8 because fixture such as `api_client` would be imported and then used as a param in test which shadows the original import. I moved fixtures to conftest.py.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,25 @@ jobs:
           when: always
           command: isort -c .
 
+  static-code-analysis-flake8:
+    docker:
+      - image: circleci/python:2.7
+        auth:
+          username: $DOCKER_USER
+          password: $DOCKER_PASS
+    working_directory: ~/code
+    steps:
+      - checkout
+
+      - run:
+          name: Prepare Environment
+          command: |
+            sudo -E pip install flake8==3.9.2
+
+      - run:
+          name: flake8
+          command: flake8
+
   build:
     machine:
       image: ubuntu-1604:201903-01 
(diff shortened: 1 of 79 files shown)


.circleci/config.yml
```diff
@@ -25,6 +25,25 @@ jobs:
           when: always
           command: isort -c .
 
+  static-code-analysis-flake8:
+    docker:
+      - image: circleci/python:2.7
```
squeaky-pl, line 30: Sadly I need a separate job with Python 2. This is of course interim solution, and hopefully would change in near future,

AlecRosenbaum:
I like the direction, but there's too much going on here for me to feel comfortable merging in one go. I don't think that I'd catch anything that's subtly different among everything else.

Can you split this up into a few different changes that first introduce the flake8 checker ignoring everything that doesn't already pass (a long todo list of ignore codes within `.flake8` would be fine), then do different PR's for each or small groups of error codes?

AlecRosenbaum (commented):
ignore -- just commenting since the above comment didn't count as a review and I don't want to get reminders

squeaky-pl:
This was addressed in other PRs.

PR #63, opened 2021-09-28
Flake8 - fix whitespace related problems

squeaky-pl:


AlecRosenbaum (approved):
just curious, does black ignore all the `bin/` files because they don't have end in `.py`?

squeaky-pl:
Yes, flake8 does the same. Since I have it set it up in my editor that's why I know about those problems. I want to actually make those files .py because this is dangerous.

AlecRosenbaum:
I think we can manually include the bin directory for flake8 at least: https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-filename

PR #64, opened 2021-09-28
Flake8 comparisons against constants

squeaky-pl:


bin/check-attachments
```diff
@@ -76,7 +76,7 @@ def main(num_accounts):
 
     with global_session_scope() as db_session:
         accounts = db_session.query(Account).filter(
-            Account.sync_should_run == True).order_by(func.rand()).limit(
+            Account.sync_should_run).order_by(func.rand()).limit(
```
squeaky-pl, line 79: Black is obviously not running here

bin/restart-forgotten-accounts
```diff
@@ -29,7 +29,7 @@ def check_accounts():
     with global_session_scope() as db_session:
         not_syncing_accounts = set(db_session.query(Account.id).
                                    filter(Account.sync_should_run,
-                                          Account.sync_host == None))  # noqa
+                                          Account.sync_host.is_(None)))
```
squeaky-pl, line 32: Neither here

bin/check-attachments
AlecRosenbaum: Is it valid to filter for `Model.field` without a condition?

tests/api/test_calendars.py
```diff
@@ -78,10 +78,8 @@ def test_delete_from_readonly_calendar(db, default_namespace, api_client):
         db.session,
         default_namespace.id,
         calendar=db.session.query(Calendar)
-        .filter(
-            Calendar.namespace_id == default_namespace.id, Calendar.read_only == True
-        )
-        .first(),  # noqa
+        .filter(Calendar.namespace_id == default_namespace.id, Calendar.read_only)
```
AlecRosenbaum, line 81: same question here about `.filter(Model.value)` without a condition

bin/check-attachments
AlecRosenbaum: Even if this is valid, if there's an option of using `.is_(True)` or == true()` (using the true expression provided by sqlalchemy) I'd prefer those.

It's not super clear to me what condition is being evaluated here if there's no obvious operator

squeaky-pl: Yes, `Model.field` in `filter` would just render `WHERE table.field` in SQL which is the same as `WHERE table.field = TRUE` AFAIK.

I agree it can be obscure to a casual reader, and typing a little bit more does not hurt as much as reading.

AlecRosenbaum (approved):

PR #66, opened 2021-09-29
Enable coverage reports

squeaky-pl:
Rendered: https://469-278426826-gh.circle-artifacts.com/0/pythoncov/index.html

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,9 +21,7 @@ services:
       dockerfile: Dockerfile
     volumes:
       - .:/opt/app
-    tmpfs:
-      - /opt/app/inbox_sync.egg-info
-      - /tmp
+    user: root
     depends_on:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -21,7 +21,7 @@ services:
       dockerfile: Dockerfile
     volumes:
       - .:/opt/app
-    tmpfs: /opt/app/inbox_sync.egg-info
+    user: root
```
squeaky-pl, line 24: Now with coverage writing to different locations it is too annoying to maintain those path mappings. Also tmpfs does not work consistently between Docker Desktop and Docker CE.

squeaky-pl: Note that the image we push does not run as root as shipped.

AlecRosenbaum (approved):

PR #68, opened 2021-09-30
Flake8: fix broken escapes

squeaky-pl:
"\s" would be first escaped on a string level, but since there is no such \s escape it's just included as it is. This is error-prone. Use r"\s" or "\\s" instead.

### DIFF
--- a/bin/inbox-start
+++ b/bin/inbox-start
@@ -38,10 +38,10 @@ from inbox.util.startup import preflight
 SOCKET_TIMEOUT = 2 * 60
 socket.setdefaulttimeout(SOCKET_TIMEOUT)
 
+esc = "\033"
 
 banner = (
-    "\033[1;95m"
-    r"""
+    r"""{esc}[1;95m
       _   _       _
      | \ | |     | |
      |  \| |_   _| | __ _ ___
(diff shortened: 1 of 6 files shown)


bin/inbox-start
```diff
@@ -39,6 +39,29 @@ SOCKET_TIMEOUT = 2 * 60
...
+    """
+    "\033[0m\033[94m"
+    """
+      S Y N C   E N G I N E
+    """
+    "\033[0m"
+    """
+     Use CTRL-C to stop.
+    """
+)
```
AlecRosenbaum, line 62: maybe this is nitpicky, but to me this is significantly less legible with it broken up like this. There's so many quotations I can't see where strings start and end without looking closely at it. Granted it's a banner, so it's not critical functionality here, but still.

Does it still pass validation if it's a single `r"""` block?

squeaky-pl: It contains terminal escapes to make it colorful and that needs `"\033"`, that would become `"\\033"` with `r""`. I came up with another idea, see if you like it.

We could also just remove color.

AlecRosenbaum (approved):

PR #74, opened 2021-10-01
Flake8: remove ambigious variable names

squeaky-pl:


tests/general/test_lock.py
```diff
@@ -8,51 +8,39 @@
...
 
 
-def lock(block, filename=None):
+def tmp_lock(block, filename=None):
     if filename is None:
         handle, filename = tempfile.mkstemp()
     return Lock(filename, block=block)
 
 
-@pytest.fixture
```
squeaky-pl, line 17: This fixtures are so simple and only used once that not having them makes code more readable

AlecRosenbaum (approved):

PR #75, opened 2021-10-04
Vendor nylas-production-python

squeaky-pl:
Too free us from the need to maintain several forks, I decided to vendor in nylas-production-python (https://github.com/nylas/nylas-production-python/tree/master/nylas)

That package contains Gunicorn worker class, structlog logging configuration, Sentry integration and a tool that spawns a background thread checking if greenlets are blocking on something (only used in gunicorn).

Refactor:

```
nylas
├── api
│   └── wsgi.py  ............ merged into inbox/api/wsgi.py
├── logging
│   ├── log.py .............. becomes inbox/logging.py
│   └── sentry .............. removed
│       └── __init__.py
└── util
    └── debug.py ............ merged into inbox/instrumentation.py
```

Additionally since we are using Rollbar I could remove raven (Sentry client) and all the supporting code. Monkeypatching nylas.logging.sentry is no longer needed as well.


bin/delete-marked-accounts
```diff
@@ -8,6 +8,8 @@ Includes:
 * Account liveness/status data (in Redis).
 
 """
+from __future__ import absolute_import
```
squeaky-pl, line 11: This is needed because on Python 2 inbox.logging would shadow stdlib logging unless absolute_import is turned on (Python 3 behavior)

inbox/api/err.py
```diff
@@ -14,15 +15,14 @@ def get_request_uid(headers):
     return headers.get("X-Unique-ID")
 
 
-def log_exception(exc_info, send_to_sentry=True, **kwargs):
```
squeaky-pl, line 17: `send_to_sentry` arg is never set to anything by callers so I just removed it.

inbox/api/wsgi.py
```diff
@@ -1,13 +1,118 @@
...
+    def __init__(self, cfg):
+        gunicorn.glogging.Logger.__init__(self, cfg)
+        configure_logging(log_level=LOGLEVEL)
+        self.error_log = log
+
 
 from inbox.config import config
 from inbox.error_handling import maybe_enable_rollbar
 
-nylas.api.wsgi.MAX_BLOCKING_TIME = config.get(
```
squeaky-pl, line 7: This was previously was done using monkeypatching. 

https://github.com/nylas/nylas-production-python/blob/master/nylas/api/wsgi.py#L13-L18

tests/api/test_sending.py
```diff
@@ -848,25 +848,8 @@ def fake_conn_pool(acct_id):
     monkeypatch.setattr("inbox.api.ns_api.writable_connection_pool", fake_conn_pool)
 
 
-@pytest.fixture
-def patch_sentry_to_raise(monkeypatch):
```
squeaky-pl, line 852: This fixture was not needed for anything, it's a false dependency of the test below

PR #84, opened 2021-10-05
Flake8: remove unused imports

squeaky-pl:


tests/auth/test_generic_auth.py
```diff
@@ -1,12 +1,11 @@
 # -*- coding: utf-8 -*-
-import copy
```
squeaky-pl, line 2: This is actually unused stdlib import I missed before

AlecRosenbaum (commented):
Is the failing test related at all?

squeaky-pl:
Need to investigate this, it seems to be consistently failing in CircleCI but not locally.

tests/contacts/test_remote_sync.py
```diff
@@ -97,7 +97,7 @@ def test_auth_error_handling(contact_sync, default_account, db):
     db.session.commit()
 
     contact_sync.start()
-    contact_sync.join(timeout=10)
+    contact_sync.join(timeout=15)
```
AlecRosenbaum, line 100: it this needed still? It looks like the test would be skipped

AlecRosenbaum (approved):

tests/contacts/test_remote_sync.py
squeaky-pl: Right, I added a separate issue about [removing it](https://github.com/closeio/sync-engine/issues/316).

squeaky-pl:
This triggered following error in `delete-accounts` pod.

Screenshot 2022-02-16 at 09 29 13


All the other pods seemed to be working correctly. Need to figure out which import has side effects of fully initializing SQLAlchemy in delete-accounts pod.

squeaky-pl:
This has rotten now, will open another one.

PR #85, opened 2021-10-06
Build arm image

froxCZ:
Arm images are built on arm machine and with `-arm64` tag suffix.

Not running tests for arm - we don't do that for any of the other services as well.


squeaky-pl (approved):
LGTM

jkemp101 (approved):

PR #88, opened 2021-10-07
Chop db shard support: part 1

squeaky-pl:
This is the minimum number of files I could safely remove to get rid off boto and munkres dependencies. They are all related to sharding.

Note that this does not get rid of all sharding code, it's just a step towards removing it. As I get familiar with the internals I will try continue this journey.


jkemp101:
We don't have any plans to support sharding and should remove it as long as it doesn't cause us more problems. @thomasst "disabled" it with https://github.com/closeio/sync-engine/commit/ef6106dfb0c490d6a67eedb2ac79a66dbc06ca9a

jkemp101 (approved):

thomasst (approved):
I believe the approach of spinning up separate clusters is better than supporting sharding.

PR #91, opened 2021-10-08
Upgrade psutil and add tests for GreenletTracer

squeaky-pl:
https://pypi.org/project/psutil/

> psutil (process and system utilities) is a cross-platform library for retrieving information on running processes and system utilization (CPU, memory, disks, network, sensors) in Python.

We are using it to monitor CPU times. The logic runs in a separate OS thread (that is not monkeypatched by gevent):

https://github.com/closeio/sync-engine/blob/b615cec56154209f3af525afd7d271b89fe2e33e/inbox/instrumentation.py#L121-L125)

Initialized here:

https://github.com/closeio/sync-engine/blob/b615cec56154209f3af525afd7d271b89fe2e33e/inbox/instrumentation.py#L106

Calculation done here:

https://github.com/closeio/sync-engine/blob/b615cec56154209f3af525afd7d271b89fe2e33e/inbox/instrumentation.py#L202-L209

This is the last version of psutil since it still works on Python 2.7.

I included lightweight test to ensure that at least things are not exploding, since previously this was not covered at all.

### DIFF
--- a/inbox/instrumentation.py
+++ b/inbox/instrumentation.py
@@ -112,13 +112,21 @@ class GreenletTracer(object):
         # thread.
         self.statsd_client = get_statsd_client()
         self.start_time = time.time()
+        self._stopping = False
+        self._thread_id = None
 
     def start(self):
         self.start_time = time.time()
         greenlet.settrace(self._trace)
         # Spawn a separate OS thread to periodically check if the active
         # greenlet on the main thread is blocking.
-        gevent._threading.start_new_thread(self._monitoring_thread, ())
+        self._thread_id = gevent._threading.start_new_thread(
+            self._monitoring_thread, ()
+        )
+
+    def stop(self):
+        assert self._thread_id, "Tracer not started"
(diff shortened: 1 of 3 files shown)


inbox/instrumentation.py
```diff
@@ -112,13 +112,21 @@ def __init__(
...
         self.start_time = time.time()
         greenlet.settrace(self._trace)
         # Spawn a separate OS thread to periodically check if the active
         # greenlet on the main thread is blocking.
-        gevent._threading.start_new_thread(self._monitoring_thread, ())
+        self._thread_id = gevent._threading.start_new_thread(
+            self._monitoring_thread, ()
+        )
+
+    def stop(self):
```
squeaky-pl, line 127: This is just to be nice and not leave this thread running for the rest of the test suite. I cannot `.join()` it because I don't have a real `threading.Thread` object, just a raw OS thread_id.

jpmelos (approved):

PR #94, opened 2021-10-09
Build Python 3 images in CI

squeaky-pl:


Dockerfile
```diff
@@ -31,8 +30,22 @@ RUN DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get dist-upgrade -y
...
   libffi-dev \
+  software-properties-common \
   && rm -rf /var/lib/apt/lists/*
 
+RUN if [ "${PYTHON_VERSION}" != "2.7" ] ; \
+  then \
+    add-apt-repository ppa:deadsnakes/ppa; \
+    DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y python"${PYTHON_VERSION}" python"${PYTHON_VERSION}"-dev; \
+  fi; \
+  if [ "${PYTHON_VERSION}" == "3.8" ] ; then DEBIAN_FRONTEND=noninteractive apt-get install python"${PYTHON_VERSION}"-distutils; fi; \
```
squeaky-pl, line 41: distutils was split into separate package on Debian somewhere after Python 3.6

Dockerfile
```diff
@@ -11,8 +12,6 @@ RUN DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get dist-upgrade -y
   g++ \
   git \
   python-dev \
-  python-pip \
```
squeaky-pl, line 14: These are now installed using pip to get the same versions on all Python versions, and not relaying on Debian packages

Dockerfile
```diff
@@ -50,12 +63,12 @@ ENV PATH="/opt/venv/bin:$PATH"
...
 COPY --chown=sync-engine:sync-engine ./ ./
 RUN \
-  virtualenv /opt/venv && \
-  pip install setuptools==44.0.0 && \
-  pip install pip==20.3.4 && \
-  pip install --no-deps -r requirements_frozen.txt && \
-  pip install -e . && \
-  pip check
+  python"${PYTHON_VERSION}" -m virtualenv /opt/venv && \
+  /opt/venv/bin/python"${PYTHON_VERSION}" -m pip install setuptools==44.0.0 pip==20.3.4 && \
```
squeaky-pl, line 67: I prefixed everything with exact Python versions to be clear that the right version of Python is invoked

Dockerfile
```diff
@@ -50,12 +63,12 @@ ENV PATH="/opt/venv/bin:$PATH"
...
+  /opt/venv/bin/python"${PYTHON_VERSION}" -m pip install --no-deps -r requirements_frozen.txt && \
+  /opt/venv/bin/python"${PYTHON_VERSION}" -m pip install -e .
+
+RUN /opt/venv/bin/python"${PYTHON_VERSION}" -m pip check
```
squeaky-pl, line 71: This step will often fail, so it made sense to have it on separate layer

.circleci/config.yml
```diff
@@ -88,6 +88,64 @@ jobs:
...
+    machine:
+      image: ubuntu-1604:201903-01
+    working_directory: ~/sync-engine
+
+    steps:
+      - checkout
+
+      - run:
+          name: build sync-engine image
+          command: docker-compose build --build-arg PYTHON_VERSION=3.6 app || true
```
jpmelos, line 102: Why do we need the `|| true` here? Shouldn't the process halt if building the image fails?

jpmelos: Oh I get it. No worries, makes total sense.

PR #96, opened 2021-10-11
Flake8: cleanup noqas

squeaky-pl:
This mostly removes dead noqas that do nothing, and fixes some problems

### DIFF
--- a/bin/unschedule-account-syncs
+++ b/bin/unschedule-account-syncs
@@ -38,7 +38,7 @@ def main(dry_run, number, hostname, process):
         if process is not None:
             hostname = ':'.join([hostname, process])
         to_unschedule = db_session.query(Account.id). \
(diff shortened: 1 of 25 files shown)


bin/unschedule-account-syncs
```diff
@@ -38,7 +38,7 @@ def main(dry_run, number, hostname, process):
         if process is not None:
             hostname = ':'.join([hostname, process])
         to_unschedule = db_session.query(Account.id). \
-                filter(Account.sync_host.like('{}%'.format(hostname)))
+            filter(Account.sync_host.like('{}%'.format(hostname)))
```
squeaky-pl, line 41: Note that this file is still not formatted with black because it does not end up in `.py`. I will be addressing this on another PR.

jpmelos (approved):

PR #102, opened 2021-10-12
HTMLParser changes for Python 3 compatibility

squeaky-pl:
HTMLParser have significant differences between Python 2 and Python 3. This is backward compatible implementation that works the same in both.

Also strengthened tests and logging while here.

### DIFF
--- a/inbox/util/html.py
+++ b/inbox/util/html.py
@@ -4,6 +4,7 @@ import re
 import sys
 
 if sys.version_info >= (3,):
+    unichr = chr
     from html.entities import name2codepoint
     from html.parser import HTMLParser
 


inbox/util/html.py
```diff
@@ -17,6 +29,8 @@ def __init__(self):
         self.fed = []
         self.strip_tag_contents_mode = False
 
+        HTMLParser.__init__(self)
```
squeaky-pl, line 32: `HTMLParser` is old style class so `super(...)` cannot be used, but this will work both in Python 2 and 3.

The explicit call to base `__init__` is needed because Python 3 introduces some changes in it, and if you don't call `__init__` the parser state will be broken.

PR #113, opened 2021-10-12
Replace rfc882 usage with email.utils

squeaky-pl:


tests/general/test_message_parsing.py
```diff
@@ -373,6 +373,16 @@ def test_address_parsing():
     parsed = parse_mimepart_address_header(mimepart, "To")
     assert parsed == [["", "alice@foocorp.com"], ["", "bob@foocorp.com"]]
 
+    # Multiple addresses on one header
+    mimepart = mime.from_string("To: alice@foocorp.com, bob@foocorp.com")
+    parsed = parse_mimepart_address_header(mimepart, "To")
+    assert parsed == [["", "alice@foocorp.com"], ["", "bob@foocorp.com"]]
+
+    # Non ASCII
```
thomasst, line 381: You can keep these tests, but both of these are already mostly covered in this one above, right?
```
    mimepart = mime.from_string(
        "To: =?utf-8?Q?Foo=2C=20Corp.?= <info@foocorp.com>, "
        "=?utf-8?Q?Support?= <support@foocorp.com>"
    )
```

inbox/util/addr.py
```diff
@@ -46,7 +46,9 @@ def parse_mimepart_address_header(mimepart, header_name):
     # Consult RFC822 Section 6.1 and RFC2047 section 5 for details.
     addresses = set()
     for section in mimepart.headers._v.getall(normalize(header_name)):
-        for phrase, addrspec in rfc822.AddressList(section).addresslist:
+        for phrase, addrspec in email.utils.getaddresses([section]):
+            if not addrspec:
```
thomasst, line 50: The behavior would change here if your `section` is `"x" <>`. As in, to maintain compatibility, you'd want this line to be `if not addrspec and not phrase:`. I'm pretty sure just skipping empty addresses as you're doing here is fine as well, though it may be better to be on the safer side here since email syncing is complex.

tests/general/test_message_parsing.py
squeaky-pl: I specifically wanted to have something that handles non ASCII, since `=2c=20` is just `, ` encoded , it's gonna become useful once I actively start testing Python 3. I will remove the case with two addresses in one header because it's not useful indeed.

PR #114, opened 2021-10-13
Log nylas_uid on HTMLParseError

squeaky-pl:
Currently we are getting large HTML snippets into the logs that confuse Loki, it also makes reading logs hard. It'd be better if we just logged some kind of id and looked up the HTML somewhere else.


AlecRosenbaum (approved):
Might be worth adding a docstring or comment to `strip_tags` that it raises `HTMLParseError` for malformed html strings.

PR #115, opened 2021-10-13
Test with S3 backend and Minio

squeaky-pl:
Minio is S3 compatible storage that can be used to easily test S3 end to end locally, https://github.com/minio/minio.

Previously we were only testing against local disk storage, S3 paths were not covered. I added a fixture that switches the backends for interesting tests and use Minio to actually store the blocks.


inbox/util/blockstore.py
```diff
@@ -11,25 +11,21 @@
...
-            h[0],
-            h[1],
-            h[2],
-            h[3],
-            h[4],
-            h[5],
-        )
 
-    def _data_file_path(h):
-        return os.path.join(_data_file_directory(h), h)
```
squeaky-pl, line 32: Previously this was conditionally enabled but I wanted to test both paths without starting separate processes. I think there is close to no value in importing this conditionally.

.circleci/config.yml
```diff
@@ -46,7 +46,7 @@ jobs:
 
   build:
     machine:
-      image: ubuntu-1604:201903-01
```
squeaky-pl, line 49: Had to update this to make the `.` in hostname work in docker-compose. This is now in par with the image we use to build ARM images.

PR #117, opened 2021-10-13
Fix sending invalid timezones through the API

jpmelos:
This should stop the API from sending invalid timezones. By invalid timezone, I mean timezones not in `pytz.all_timezones`. The timezones in the value-side of the new timezone table added in this PR are all recognized by pytz.

### DIFF
--- a/inbox/api/kellogs.py
+++ b/inbox/api/kellogs.py
@@ -86,7 +86,7 @@ def _convert_timezone_to_iana_tz(original_tz):
     try:
         return timezones_table[original_tz]
     except KeyError:
-        log.error("Bad timezone idetifier", original_tz=original_tz)
+        log.error("Bad IANA timezone identifier", original_tz=original_tz)
         return original_tz
 
 


jpmelos:
@squeaky-pl FYI @jkemp101 asked that we wait for @thomasst's review before shipping this.

inbox/api/kellogs.py
```diff
@@ -81,6 +82,15 @@ def encode(obj, namespace_public_id=None, expand=False, is_n1=False):
         raise
 
 
+def _convert_timezone_to_pytz(original_tz):
+    try:
+        return timezones_table[original_tz]
+    except KeyError:
+        # Should we log a rollbar here so someone can come in the table and add
+        # the value?
```
thomasst, line 90: We should log it (as a warning or error).

inbox/api/kellogs.py
```diff
@@ -81,6 +82,15 @@ def encode(obj, namespace_public_id=None, expand=False, is_n1=False):
         raise
 
 
+def _convert_timezone_to_pytz(original_tz):
```
thomasst, line 85: Probably just `_convert_timezone_to_tz` since it has nothing to do with pytz? I believe it's called the "tz database".

thomasst (approved):

PR #118, opened 2021-10-14
Remove liblua

squeaky-pl:
This no longer seems necesarry after: https://github.com/closeio/sync-engine/pull/111


froxCZ (approved):
Also see it in setup.sh ... not sure if that is used for anything though

squeaky-pl:
@froxCZ that is part of Vagrant setup that eventually will be remove as well

PR #121, opened 2021-10-18
Fix tests/security/test_blobstorage.py Python 3 compatibility

squeaky-pl:


tests/security/test_blobstorage.py
```diff
@@ -17,7 +17,7 @@ def test_blobstorage(config, sample_input, encrypt):
 def test_encoded_format(config, sample_input, encrypt):
     config["ENCRYPT_SECRETS"] = encrypt
     encoded = encode_blob(sample_input)
-    assert encoded.startswith(chr(encrypt) + "\x00\x00\x00\x00")
```
squeaky-pl, line 20: This is a cryptic way of returning `"\x01"` and `"\x00"`. Since encrypt is `bool` and in Python `bool` is a subclass of `int`, `True` becomes `1` and `False` becomes `0`.

tests/security/test_blobstorage.py
```diff
@@ -17,7 +17,7 @@ def test_blobstorage(config, sample_input, encrypt):
...
+    assert encoded.startswith((b"\x01" if encrypt else b"\x00") + b"\x00\x00\x00\x00")
```
squeaky-pl, line 20: We want to work with bytes here both in Python 2 and Python 3 since this is binary data.

PR #122, opened 2021-10-18
Fix tests/api/test_files.py for Python 3 compatibility

squeaky-pl:


requirements_frozen.txt
```diff
@@ -37,6 +37,7 @@ Flask==0.10.1
 Flask-RESTful==0.3.2
 freezegun==0.3.7
 funcsigs==1.0.2
+future==0.18.2
```
squeaky-pl, line 40: This provides builtins module on Python 2

inbox/util/encoding.py
```diff
@@ -23,6 +27,9 @@ def unicode_safe_truncate(s, max_length):
     Implements unicode-safe truncation and trims whitespace for a given input
     string, number or unicode string.
     """
-    if not isinstance(s, unicode):
-        s = str(s).decode("utf-8", "ignore")
+    if isinstance(s, bytes):
```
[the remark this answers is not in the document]
squeaky-pl: The logic needs to be split to extra branch because on Python 3 you cannot call `.decode()` on Python 3 `str`.

tests/api/test_files.py
```diff
@@ -142,7 +143,9 @@ def test_download(api_client, uploaded_file_ids, filename):
         os.path.dirname(os.path.abspath(__file__)),
         "..",
         "data",
-        original_filename.encode("utf-8"),
+        original_filename.encode("utf-8")
```
squeaky-pl, line 146: * on Python 2 the programmer is responsible for encoding non ASCII file names into the file system encoding.
* on Python 3 this is taken care of underlying Python stdlib, only `str` instances are accepted, not `bytes`.

PR #123, opened 2021-10-18
Fix inbox/util/file.py Python 3 compatibility

squeaky-pl:


inbox/util/file.py
```diff
@@ -2,9 +2,19 @@
...
 import os
 import string
+import sys
 
 from gevent.coros import BoundedSemaphore
 
+if sys.version_info < (3,):
+    import builtins
+
+    file_like = builtins.file
```
squeaky-pl, line 12: on Python 2 the "base type" for files is `file`

PR #126, opened 2021-10-19
Fix tests/security/test_secret.py for Python 3 compatibility

squeaky-pl:


tests/security/test_secret.py
```diff
@@ -78,11 +81,14 @@ def test_token(db, config, encrypt):
     if encrypt:
         assert secret._secret != token, "token not encrypted"
     else:
-        assert secret._secret == token, "token encrypted when encryption disabled"
+        assert secret._secret == token.encode(
```
squeaky-pl, line 84: on Python 3 `bytes` cannot be compared with `str` so you need explicit `encode()`

PR #138, opened 2021-10-21
Fix exception variable leaking to outer scope in Python 2 only

squeaky-pl:


inbox/sendmail/smtp/postel.py
```diff
@@ -337,10 +338,12 @@ def _send(self, recipients, msg):
                             failures=failures,
                         )
             except smtplib.SMTPException as err:
+                last_error = err
```
squeaky-pl, line 341: To make it work in Python 2 and 3 I introduce an extra variable and store the last exception here.

PR #140, opened 2021-10-21
Convert filter and map to comprehensions

squeaky-pl:
This works both on Python 2 & 3

### DIFF
--- a/inbox/events/recurring.py
+++ b/inbox/events/recurring.py
@@ -144,8 +144,9 @@ def get_start_times(event, start=None, end=None):
 
             # We want naive-everything for all-day events.
             if event.all_day:
-                excl_dates = map(lambda x: x.naive, excl_dates)
-            map(rrules.exdate, excl_dates)
+                excl_dates = [x.naive for x in excl_dates]
+            for excl_date in excl_dates:
+                rrules.exdate(excl_date)
 
         # Return all start times between start and end, including start and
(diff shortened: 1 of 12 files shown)


inbox/events/recurring.py
```diff
@@ -144,8 +144,9 @@ def get_start_times(event, start=None, end=None):
 
             # We want naive-everything for all-day events.
             if event.all_day:
-                excl_dates = map(lambda x: x.naive, excl_dates)
-            map(rrules.exdate, excl_dates)
+                excl_dates = [x.naive for x in excl_dates]
+            for excl_date in excl_dates:
```
squeaky-pl, line 148: Here map was executed for side effects. It's more readable as explicit loop.

inbox/util/itert.py
```diff
@@ -24,7 +24,7 @@ def partition(pred, iterable):
             partition(is_odd, range(10)) --> 0 2 4 6 8   and  1 3 5 7 9
     """
     t1, t2 = itertools.tee(iterable)
-    return list(itertools.ifilterfalse(pred, t1)), filter(pred, t2)
+    return list(itertools.ifilterfalse(pred, t1)), [e for e in t2 if pred(e)]
```
jpmelos, line 27: Is it also possible to turn `list(itertools.ifilterfalse(pred, t1))` into a list comp?

PR #169, opened 2021-11-03
Upgrade boto3 & botocore

squeaky-pl:
Boto3 & botocore are only used in contacts sync which we don't use. Those are the latest versions that still work on Python 2.


requirements_frozen.txt
```diff
@@ -109,6 +109,7 @@ redis==2.10.6
 regex==2018.1.10
 requests==2.26.0
 rollbar==0.16.1
+s3transfer==0.4.2
```
squeaky-pl, line 112: This is new dependency of boto3

PR #192, opened 2021-11-05
Upgrade imapclient

squeaky-pl:
This is the same version of imapclient we use in another project of ours.

The fixes we've made have been independently implemented upstream and are included in the latest release.
I've run manual email syncing tests with this version of imapclient.

### DIFF
--- a/requirements_frozen.txt
+++ b/requirements_frozen.txt
@@ -53,7 +53,7 @@ html2text==2019.8.11
 hypothesis==4.57.1
 icalendar==3.9.1
 idna==2.9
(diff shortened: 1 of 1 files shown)


jpmelos (approved):
Do we have other projects that use our fork? If so, could we also upgrade them to upstream? How much effort would that be? And if/when we don't have projects using our fork, should we kill it?

squeaky-pl:
@jpmelos  We definitely should archive the fork, that's what happened to our old flanker fork.

sync-engine is the only one that still uses this fork: https://github.com/search?q=org%3Acloseio+imapclient.git&type=code

PR #193, opened 2021-11-08
refresh_token is str

squeaky-pl:
This is very analogical to: https://github.com/closeio/sync-engine/pull/183

Refresh tokens are stored in MySQL in binary BLOB fields.

The logic already accepts bytes and text, and if it receives text it will encode it UTF-8 before saving.

On Python 3 the getters will return bytes. This is fine on Python 2 where it is a synonym of string and silently coerces to unicode in comparisons. But Python 3 needs an extra .decode('utf-8').

### DIFF
--- a/tests/security/test_secret.py
+++ b/tests/security/test_secret.py
@@ -1,7 +1,4 @@
 # -*- coding: UTF-8 -*-
-import builtins
-import sys
-
 import pytest
 
 from inbox.auth.google import GoogleAccountData, GoogleAuthHandler
@@ -85,10 +82,11 @@ def test_token(db, config, encrypt):
             "utf-8"
         ), "token encrypted when encryption disabled"
 
-    decrypted_secret = secret.secret
-    assert (
-        decrypted_secret == token.encode("utf-8")
-        and account.refresh_token == decrypted_secret
+    decrypted_secret = secret.secret  # type: bytes
+    assert decrypted_secret == token.encode(
+        "utf-8"
+    ) and account.refresh_token == decrypted_secret.decode(
+        "utf-8"
     ), "token not decrypted correctly"
 
     # db.session.delete(account.auth_credentials[0])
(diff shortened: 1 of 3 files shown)


tests/security/test_secret.py
```diff
@@ -118,21 +116,15 @@ def test_token_inputs(db, config, encrypt, default_account):
...
+    assert isinstance(secret.secret, bytes), "secret cannot be unicode"
     assert secret.secret == unicode_token.encode(
         "utf-8"
     ), "token not decrypted correctly"
 
-    with pytest.raises(ValueError) as e:
+    with pytest.raises(ValueError):
         default_account.refresh_token = invalid_token
 
-    assert e.typename == "ValueError", "token cannot be invalid UTF-8"
```
squeaky-pl, line 131: This was not testing anything because it's already asserted that it's ValueError in pytest.raises

tests/security/test_secret.py
```diff
@@ -118,21 +116,15 @@ def test_token_inputs(db, config, encrypt, default_account):
...
-
-    with pytest.raises(ValueError) as f:
+    with pytest.raises(ValueError):
         default_account.refresh_token = null_token
 
-    assert f.typename == "ValueError", "token cannot contain NULL byte"
```
squeaky-pl, line 136: Same as above

tests/security/test_secret.py
```diff
@@ -118,21 +116,15 @@ def test_token_inputs(db, config, encrypt, default_account):
     secret_id = default_account.refresh_token_id
     secret = db.session.query(Secret).get(secret_id)
 
-    assert not isinstance(
-        secret.secret, builtins.unicode if sys.version_info < (3,) else str
-    ), "secret cannot be unicode"
+    assert isinstance(secret.secret, bytes), "secret cannot be unicode"
```
squeaky-pl, line 119: this is just easier to reason on both Python 2 & 3

jpmelos (approved):

PR #195, opened 2021-11-08
CONDSTORE related bytes/str changes

squeaky-pl:
This is a couple of str/bytes changes around the code that uses `CONDSTORE`.

`CONDSTORE` is a capability that allows the server to ask what messages changed in given folder since given `MODSEQ` (modification sequence). At least gmail supports it.

Modification sequences are monotonically increasing integers that server assigns to each message on change. If you ask the server for a `HIGESTMODSEQ` (highest modification sequence) in given folder, then sleep for a while, then `FETCH` messages since that highest modification sequence, you can implement efficient way to sync incremental changes, without `FETCH`ing all the messages from scratch.

### DIFF
--- a/inbox/util/testutils.py
+++ b/inbox/util/testutils.py
@@ -218,12 +218,12 @@ class MockIMAPClient(object):
                 m = re.match("CHANGEDSINCE (?P<modseq>[0-9]+)", modifiers[0])
                 if m:
                     modseq = int(m.group("modseq"))
-                    items = {u for u in items if uid_dict[u]["MODSEQ"][0] > modseq}
+                    items = {u for u in items if uid_dict[u][b"MODSEQ"][0] > modseq}
         data = [d.encode() for d in data]
         for u in items:
             if u in uid_dict:
                 resp[u] = {
-                    k: v for k, v in uid_dict[u].items() if k in data or k == "MODSEQ"
+                    k: v for k, v in uid_dict[u].items() if k in data or k == b"MODSEQ"
                 }
         return resp
 
@@ -257,7 +257,10 @@ class MockIMAPClient(object):
         lastuid = max(folder_data) if folder_data else 0
         resp = {b"UIDNEXT": lastuid + 1, b"UIDVALIDITY": self.uidvalidity}
         if data and "HIGHESTMODSEQ" in data:
-            resp[b"HIGHESTMODSEQ"] = max(v[b"MODSEQ"] for v in folder_data.values())
+            resp[b"HIGHESTMODSEQ"] = max(
+                v[b"MODSEQ"][0] if isinstance(v[b"MODSEQ"], tuple) else v[b"MODSEQ"]
+                for v in folder_data.values()
+            )
         return resp
 
     def delete_messages(self, uids, silent=False):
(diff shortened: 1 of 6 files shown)


inbox/util/testutils.py
```diff
@@ -257,7 +257,7 @@ def folder_status(self, folder_name, data=None):
         lastuid = max(folder_data) if folder_data else 0
         resp = {b"UIDNEXT": lastuid + 1, b"UIDVALIDITY": self.uidvalidity}
         if data and "HIGHESTMODSEQ" in data:
-            resp[b"HIGHESTMODSEQ"] = max(v[b"MODSEQ"] for v in folder_data.values())
+            resp[b"HIGHESTMODSEQ"] = max(v[b"MODSEQ"][0] for v in folder_data.values())
```
squeaky-pl, line 260: This area setups fake data for the tests, we don't use this code in real syncs.

`MODSEQ` as stored in each message is always returned as a one element tuple containing an integer e.g. `(123,)` on the protocol level.

The `HIGESTMODSEQ` is a single integer, so it should be 123.

Modification sequences are compared in code as in a > b, means that a is newer than b.

There was an error here since forever that was hidden because of sloppiness of Python 2:

```python
>>> 1 < (1,)
True
```

While on Python 3:

```python
>>> 1 < (1,)
  File "<stdin>", line 1, in <module>
TypeError: '<' not supported between instances of 'int' and 'tuple'
``` 

All in all the two tests that rely on this data worked by mistake in Python 2...

tests/imap/data.py
```diff
@@ -39,7 +39,7 @@ def build_uid_data(internaldate, flags, body, g_labels, g_msgid, modseq):
         b"X-GM-LABELS": g_labels,
         b"X-GM-MSGID": g_msgid,
         b"X-GM-THRID": g_msgid,  # For simplicity
-        b"MODSEQ": modseq,
+        b"MODSEQ": (modseq,),
```
squeaky-pl, line 42: Also here generate what actually happens in real life

tests/imap/test_delete_handling.py
```diff
@@ -301,9 +301,9 @@ def test_renamed_label_refresh(
 
     new_flags = {
         msg_uid: {
-            "FLAGS": ("\\Seen",),
-            "X-GM-LABELS": ("new label",),
-            "MODSEQ": ("23",),
+            b"FLAGS": (b"\\Seen",),
+            b"X-GM-LABELS": (b"new label",),
+            b"MODSEQ": (23,),
```
squeaky-pl, line 306: This tests Gmail, in Gmail modseqs are always integers, so simulate real life better.

Note that generic IMAP can return this as something, but we always normalize it somewhere in the code.

tests/imap/test_folder_sync.py
```diff
@@ -127,6 +127,9 @@ def test_condstore_flags_refresh(
         v[b"X-GM-LABELS"] = (b"newlabel",)
         v[b"MODSEQ"] = (k,)
 
+    with folder_sync_engine.conn_pool.get() as crispin_client:
```
squeaky-pl, line 130: Wanted this checked to confirm that this test actually uses CONDSTORE.

jpmelos (approved):

PR #198, opened 2021-11-09
Fix SMTP host

squeaky-pl:
Newer versions of Python verify that the host we are conecting to matches the cert, one needs to provide host to the constructor, otherwise it wil fail.

We don't use this part of code (sending) but it was easy to fix anyway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,7 +62,6 @@ ENV PATH="/opt/venv/bin:$PATH"
 
 COPY --chown=sync-engine:sync-engine ./ ./
 RUN \
-  --mount=type=cache,target=/home/sync-engine/.cache,uid=5000,gid=5000  \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -62,6 +62,7 @@ ENV PATH="/opt/venv/bin:$PATH"
 
 COPY --chown=sync-engine:sync-engine ./ ./
 RUN \
+  --mount=type=cache,target=/home/sync-engine/.cache,uid=5000,gid=5000  \
```
jpmelos, line 65: I'm guessing this one doesn't belong here?

jpmelos (approved):

Dockerfile
squeaky-pl: Yeah, haha. I use this to speed up repetitve installs locally but I didn't mean to commit it haha.

PR #201, opened 2021-11-09
Fix and filter some warnings

squeaky-pl:
This cleanups some warnings.

We have some warnings related to mysqlclient that are very spammy but we cannot do anything about them now since we are stuck on a version from 2018, which is the last version that cooperates with gevent.


inbox/api/ns_api.py
```diff
@@ -19,7 +19,7 @@
     request,
     stream_with_context,
 )
-from flask.ext.restful import reqparse
```
squeaky-pl, line 22: That packages name is deprecated

jpmelos (approved):

PR #208, opened 2021-11-11
SSL changes for Python 3

squeaky-pl:
This gates all the SSL fixes to make SSL decent on  Python 2 conditionally, we don't need them on Python 3 since they are already there in Python 3 ssl module.

Once we sunset Python 2 I am gonna remove all the Python 2 branches.

### DIFF
--- a/bin/contact-search-backfill.py
+++ b/bin/contact-search-backfill.py
@@ -1,11 +1,14 @@
 #!/usr/bin/env python
+import sys
+
 from gevent import monkey
 
 monkey.patch_all()
 
-import gevent_openssl
+if sys.version_info < (3,):
+    import gevent_openssl
 
-gevent_openssl.monkey_patch()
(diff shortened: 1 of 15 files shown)


bin/contact-search-backfill.py
```diff
@@ -1,11 +1,14 @@
 #!/usr/bin/env python
+import sys
```
jpmelos, line 2: As a general rule, I believe every import should come after `gevent`'s `monkey.patch_all()`. This is explicitly mentioned in `gevent`'s docs: https://www.gevent.org/api/gevent.monkey.html.

So I think these files should start with:

```python
from gevent import monkey

monkey.patch_all()

import sys
...
```

squeaky-pl: @jpmelos Right, fixed!

jpmelos (approved):

PR #210, opened 2021-11-11
Upgrade sqlalchemy

squeaky-pl:
These are the remaining changes needed to make sync-engine work with SQLAlchemy 1.4.

We used a fork of sqlalchemy because original authors tweaked so called baked query subsystem to control how much memory it uses.

See: https://github.com/closeio/sync-engine/pull/211


inbox/models/thread.py
```diff
@@ -228,7 +228,10 @@ def mark_for_deletion(self):
         self.deleted_at = datetime.datetime.utcnow()
 
     discriminator = Column("type", String(16))
-    __mapper_args__ = {"polymorphic_on": discriminator}
+    __mapper_args__ = {
+        "polymorphic_on": discriminator,
+        "polymorphic_identity": "thread",
```
squeaky-pl, line 233: newer SQLAlchemy versions required polymorphic_identity to be specified even on the base classes with SQLAlchemy inheritance. We never use the base polymorphic_identity value ourselves so the change from missing to `thread` does not affect us.

jpmelos (approved):

PR #213, opened 2021-11-11
SQLAlchemy cache_ok

squeaky-pl:
This marks type decorators as ready for SQLAlchemy query caching mechanism.


jpmelos (approved):
Just to make sure I understand this right, `cache_ok = True` means "this type will always generate the same SQL given the same parameters and bound value", right? As in, it's a stateless serializer and deserializer (except for the state of the initial instantiation of the type).

squeaky-pl:
@jpmelos yes, it basically means that those functions are "pure".

PR #217, opened 2021-11-12
Disable JSON_SORT_KEYS

squeaky-pl:
I've stumbled on mysterious

`TypeError: '<' not supported between instances of 'NoneType' and 'str'`

in `json.dumps` when manually testing things on Python 3.6.

It turns out Python 3.6 has a bug when one passes `sort_keys=True` into `json.dumps` and dictionaries have `None` keys and at least syncback uses such dictionaries. We don't pass it ourselves but we are using it indirectly with Flask's `jsonify`. I had to disable it on 3.6. We are going for 3.8 anyway so I think this is acceptable intermediate step.

See also: https://github.com/humanprotocol/hmt-escrow/issues/32

### DIFF
--- a/inbox/api/srv.py
+++ b/inbox/api/srv.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import
 
+import sys
+
 from flask import Flask, g, jsonify, make_response, request
 from flask_restful import reqparse
 from sqlalchemy.orm.exc import NoResultFound
@@ -31,6 +33,8 @@ from .metrics_api import app as metrics_api
 from .ns_api import DEFAULT_LIMIT, app as ns_api
 
 app = Flask(__name__)
+if (3, 6) <= sys.version_info < (3, 7):
+    app.config["JSON_SORT_KEYS"] = False
 # Handle both /endpoint and /endpoint/ without redirecting.
 # Note that we need to set this *before* registering the blueprint.
 app.url_map.strict_slashes = False
--- a/inbox/api/srv.py
+++ b/inbox/api/srv.py
@@ -1,7 +1,5 @@
 from __future__ import absolute_import
 
-import sys
-
 from flask import Flask, g, jsonify, make_response, request
 from flask_restful import reqparse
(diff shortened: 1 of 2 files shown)


jpmelos:
I wouldn't call it a "bug". I mean, you *are* trying to compare uncomparable types. How is Python supposed to sort a dictionary that has uncomparable types mixed?

Also, you're hedging against this only for `3.6 <= version < 3.7`, but I see the same behavior in Python 3.8:

```
Python 3.8.11 (default, Jul  6 2021, 10:50:43) 
[Clang 12.0.5 (clang-1205.0.22.11)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> json.dumps({'a': 1, 'b': 2})
'{"a": 1, "b": 2}'
>>> json.dumps({'a': 1, 'b': 2}, sort_keys=True)
'{"a": 1, "b": 2}'
>>> json.dumps({'a': 1, 'b': 2, None: 3}, sort_keys=True)
  File "<stdin>", line 1, in <module>
  File "/Users/jpmelos/.pyenv/versions/3.8.11/lib/python3.8/json/__init__.py", line 234, in dumps
    return cls(
  File "/Users/jpmelos/.pyenv/versions/3.8.11/lib/python3.8/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/Users/jpmelos/.pyenv/versions/3.8.11/lib/python3.8/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
TypeError: '<' not supported between instances of 'NoneType' and 'str'
```

Did you manage to make it work for Python 3.7+? If not, I'd say we should just disable it entirely. Do you agree?

Anyway, I guess I wouldn't consider disabling JSON sort keys here as a case of "working around a bug", but instead fixing the `sync-engine` code...

squeaky-pl:
@jpmelos OK, let's disable it altogether.

I only tested on 3.6 and I took what was written in the other discussion for granted. Didn't test 3.8.

jpmelos (approved):

PR #221, opened 2021-11-15
Gmail labels containing asterisks search fix

squeaky-pl:
Not related to Python 3 porting, just a problem we had before.

We are sporadically hitting errors when customers have labels in their Gmail inbox containing asterisks. Such us:

```
InvalidCriteriaError: b'Could not parse command'
``` 

For an IMAP search criteria like `["X-GM-LABELS", "*Asterisk"]`

This is an extension to IMAP by Google and we are using it. It also works differently than the rest of IMAP as implemented by `imapclient` library, specifically:
* the label has to be utf-7 encoded (https://en.wikipedia.org/wiki/UTF-7). This is documented here: https://developers.google.com/gmail/imap/imap-extensions 
* if the label contains asterisk it has to be quoted

We were already doing the first thing already but in the wrong place architecture-wise. 
About quoting ,`imapclient` only quotes values containing `"` and `\` characters and leaves everything else unquoted. This requirement to quote labels containing `*` character is not documented anywhere, I have found out it has to be done by experimenting. I don't think IMAP RFC mandates this, although `*` has special meaning in some contexts/positions but here it would not be ambiguous. This is most probably Gmail quirk, whilst `imapclient` follows the RFC.

I've also confirmed this works by creating a Gmail label containing *Asterisk in my Gmail inbox and then used the code from this pull request. It would fail loudly before.

Quoting labels not containing `*` is not necessary, but quoting them always does not hurt either. This is a little bit hacky because `imapclient`s default logic has to be skipped.

### DIFF
--- a/inbox/crispin.py
+++ b/inbox/crispin.py
@@ -1390,64 +1390,60 @@ class GmailCrispinClient(CrispinClient):
     def search_uids(self, criteria):
         # type: (List[str]) -> List[int]
         """
-        Find UIDs in this folder matching the criteria. See
-        http://tools.ietf.org/html/rfc3501.html#section-6.4.4 for valid
-        criteria.
+        Handle Gmail label search oddities.
+        https://developers.google.com/gmail/imap/imap-extensions#access_to_gmail_labels_x-gm-labels.
+
+        UTF-7 encodes label names and also quotes it to prevent errors when the label contains
+        asterisks (*). imapclient's search method sends label names containing asterisks unquoted which
+        upsets Gmail IMAP server.
         """
-        if len(criteria) == 2 and criteria[0] == "X-GM-LABELS":
-            response = gmail_label_search(self.conn, criteria[1])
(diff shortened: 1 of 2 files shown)


inbox/crispin.py
```diff
@@ -1382,3 +1386,68 @@ def delete_sent_message(self, message_id_header, delete_multiple=False):
...
         return True
+
+    def search_uids(self, criteria):
+        # type: (List[str]) -> List[int]
+        """
+        Find UIDs in this folder matching the criteria. See
+        http://tools.ietf.org/html/rfc3501.html#section-6.4.4 for valid
+        criteria.
+        """
+        if len(criteria) == 2 and criteria[0] == "X-GM-LABELS":
```
squeaky-pl, line 1397: This just hijacks the flow here to call special version of the logic.

jpmelos (approved):

PR #236, opened 2021-11-19
Skip tests

squeaky-pl:
Provide ways to skip tests using #notests similary to other repositories.This is sometimes useful when reverting but should be used with care of course.


jpmelos (approved):
Had forgotten this one. 🤦 Sorry!

PR #253, opened 2021-11-25
Fix for contact name being list

squeaky-pl:
The underlying library sometimes parses name as a list. From the observations in Rollbar we know that the first element is the actual name and the second one is a title.


drewler (approved):
Approved. Would be nice to see a Rollbar for extra context.

PR #262, opened 2021-11-26
Upgrade regex

squeaky-pl:
This is flanker dependency.

https://github.com/mrabarnett/mrab-regex does not provide a changelog or tags so ...
It's mostly extending stdlib re and adding new features only so it should be safe to do.

### DIFF
--- a/requirements/prod.txt
+++ b/requirements/prod.txt
@@ -92,7 +92,8 @@ python-dateutil==2.8.2
 pytz==2021.3
 PyYAML==5.4.1
 redis==2.10.6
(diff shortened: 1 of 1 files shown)


drewler (approved):
👍 

Out of curiosity, I was looking if there's a way to get all the issues/PRs closed within a date range using Github's search, as a poor man's substitute for a changelog. Found this, in case it could be useful in the future:

https://github.com/mrabarnett/mrab-regex/issues?q=closed%3A2021-03-17..2021-11-10

PR #266, opened 2021-11-29
Remove colorama

squeaky-pl:
Nothing depends on it


drewler (approved):
👍 

`pipdeptree -p` and `pipdeptree -rp` outputs would be nice for extra context (same applies to the other dependency-dropping PRs)

PR #285, opened 2021-12-01
Flake8-simplify

squeaky-pl:
The ignore configuration is the same like in our other projects, except a couple of extra codes because we still need to keep Python 2 compatibility.

We plan to drop Python 2.7 in first days of C1. I created an issue: https://github.com/closeio/sync-engine/issues/286

### DIFF
--- a/inbox/api/update.py
+++ b/inbox/api/update.py
@@ -220,52 +220,46 @@ def update_message_labels(
         else:
             removed_labels.append(category.display_name)
 
-    # XXX: Non-optimistic updates are buggy because if we don't update the
-    # state and can't supply added/removed labels we'll end up with
-    # inconsistencies.
-    if optimistic:
-        # Use a consistent time across creating categories, message updated_at
-        # and the subsequent transaction that may be created.
-        update_time = datetime.utcnow()
+    # Use a consistent time across creating categories, message updated_at
+    # and the subsequent transaction that may be created.
+    update_time = datetime.utcnow()
 
-        # Optimistically update message state,
-        # in a manner consistent with Gmail.
-        for cat in added_categories:
-            # message.categories.add(cat)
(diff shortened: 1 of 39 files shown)


inbox/api/update.py
```diff
@@ -223,7 +223,7 @@ def update_message_labels(
     # XXX: Non-optimistic updates are buggy because if we don't update the
     # state and can't supply added/removed labels we'll end up with
     # inconsistencies.
-    if optimistic or True:
```
jpmelos, line 226: LOL what?

inbox/api/update.py
```diff
@@ -223,7 +223,7 @@ def update_message_labels(
...
+    if optimistic:
```
jpmelos, line 226: This branch would always run, and now it won't, right? Is this the intention here? I would expect this PR to mostly not change behavior, that's why I'm checking explicitly that this is what is intended here.

inbox/mailsync/backends/imap/generic.py
```diff
@@ -445,12 +445,11 @@ def initial_sync_impl(self, crispin_client):
             bind_context(change_poller, "changepoller", self.account_id, self.folder_id)
             uids = sorted(new_uids, reverse=True)
             count = 0
```
jpmelos, line 447: Can this line be dropped?

inbox/api/update.py
squeaky-pl: True dat. Thanks for catching this.

I looked at the code a little bit to understand what `optimistic` is, and it looks like it's a runtime feature flag triggered by client HTTP API header. I need to double check but it looks like we would always want optimistic behavior in other places of code as well because we don't set that header ourselves.

jpmelos (approved):

PR #292, opened 2021-12-02
Fix plaintext2html

squeaky-pl:
Out first Python 3.8 bug.

`cgi.escape` was deprecated since 3.2 (https://docs.python.org/3.7/library/cgi.html#cgi.escape) and was finally removed in 3.8.  One is suppoed to use `html.escape`. Note that there is a difference between `quote` param which is True by default in `html.escape`. We will be generating things that don't end up in HTML attributes so we can leave it set to False in both cases.

### DIFF
--- a/inbox/util/html.py
+++ b/inbox/util/html.py
@@ -113,7 +113,7 @@ def plaintext2html(text, tabstop=4):
     def do_sub(m):
         c = m.groupdict()
         if c["htmlchars"]:
-            return html_escape(c["htmlchars"], False)
+            return html_escape(c["htmlchars"], quote=False)
         if c["lineend"]:
             return "<br>"
         elif c["space"]:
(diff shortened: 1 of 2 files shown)


inbox/util/html.py
```diff
@@ -96,7 +101,7 @@ def strip_tags(html):
 
 # https://djangosnippets.org/snippets/19/
 re_string = re.compile(
-    r"(?P<htmlchars>[<&>])|(?P<space>^[ \t]+)|(?P<lineend>\n)|(?P<protocol>(^|\s)((http|ftp)://.*?))(\s|$)",
+    r"(?P<htmlchars>[<&>])|(?P<space>^[ \t]+)|(?P<lineend>\n)|(?P<protocol>(^|\s)((https?|ftp)://.*?))(\s|$)",
```
squeaky-pl, line 104: in 2021 we can also handle HTTPs ;-)

jpmelos (approved):

PR #294, opened 2021-12-06
Remove percent formatting aka flake8-sfs

squeaky-pl:
Many of those would benefit from fstrings but we can't until January when we plan to drop Python 2.7.

Settings from other project of ours. I left an issue for reviewing places where fstrings would simplify things: https://github.com/closeio/sync-engine/issues/296

### DIFF
--- a/tests/system/test_sending.py
+++ b/tests/system/test_sending.py
@@ -48,7 +48,7 @@ def wait_for_trash(client, thread_id):
 @pytest.mark.parametrize("client", all_accounts)
 def test_sending(client):
     # Create a message and send it to ourselves
-    subject = "%s (Self Send Test)" % strftime("%Y-%m-%d %H:%M:%S")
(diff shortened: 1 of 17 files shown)


tests/system/test_sending.py
```diff
@@ -48,7 +48,7 @@ def wait_for_trash(client, thread_id):
 @pytest.mark.parametrize("client", all_accounts)
 def test_sending(client):
     # Create a message and send it to ourselves
-    subject = "%s (Self Send Test)" % strftime("%Y-%m-%d %H:%M:%S")
+    subject = "{} (Self Send Test)".format(strftime("%Y-%m-%d %H:%M:%S"))
```
squeaky-pl, line 51: Note that this not make much sense because strftime needs more arguments but those tests are not even run... They are broken and they were ignored in pytest.ini before we even started maintaining this project. They will most probably be chopped at some point, but on a separate PR/discussion.

jpmelos (approved):

PR #303, opened 2021-12-29
Message parsing length limit

thomasst:
Fixes #

Summary:

Test Plan:

Reviewers:


jkemp101:
My thoughts,

* I think we should try to do the check before saving to S3 [here](https://github.com/closeio/sync-engine/blob/838d8660191bccbf70e19e52b4014934aef174a1/inbox/models/message.py#L298) so we aren't left with huge orphaned S3 objects.
* Add a constant in [here](https://github.com/closeio/sync-engine/blob/master/inbox/models/constants.py) for max message size?
* We should probably add a test for this to confirm the behavior.

thomasst:
> * I think we should try to do the check before saving to S3 [here](https://github.com/closeio/sync-engine/blob/838d8660191bccbf70e19e52b4014934aef174a1/inbox/models/message.py#L298) so we aren't left with huge orphaned S3 objects.

Why orphaned? Wouldn't we still want to store the RFC822 original even if we can't parse the message?

jkemp101 (approved):

PR #304, opened 2022-01-04
Fix running tests in Docker on M1

thomasst:
- The `mysql` image doesn't have M1 support, so I changed it to `mysql/mysql-server`, and also to use the latest version.
- By default, you can only use the root password from within the container, so I had to add a `MYSQL_ROOT_HOST` setting to allow connecting from anywhere. Ideally, we wouldn't use the root password.

I was then able to run the tests locally. But it was still using Python 2.7.
```
docker-compose run -e 'NYLAS_ENV=test' app pytest
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,9 +6,10 @@ services:
       - 6379
 
   mysql:
-    image: mysql:5.7.31
+    image: mysql/mysql-server:8.0.27
     environment:
       MYSQL_ROOT_PASSWORD: nylas_it
+      MYSQL_ROOT_HOST: "%"
       MYSQL_USER: inboxtest
       MYSQL_PASSWORD: inboxtest
       MYSQL_DATABASE: synctest


thomasst:
@squeaky-pl Can you figure out how we can test locally using Py3 as well? And we should also add this to the README.

thomasst:
On the latest master it's now using Python 3.8 to run the tests, but it fails for me:

```
../venv/lib/python3.8/site-packages/_pytest/assertion/rewrite.py:170: in exec_module
    exec(co, module.__dict__)
tests/system/conftest.py:20: in <module>
    from .google_auth_helper import google_auth
tests/system/google_auth_helper.py:5: in <module>
    from inbox.auth.gmail import GmailAuthHandler
E   ModuleNotFoundError: No module named 'inbox.auth.gmail'
```

squeaky-pl:
I have a small doubt here about changing the version of MySQL, this could be potentially dangerous if somebody writes code that uses newer features of MySQL and then we don't catch this. Maybe my worries are overblown.

One solution to this is running sed in CircleCI to replace it back to the old version on x86_64.

What do you think @AlecRosenbaum @jkemp101

jkemp101:
I think this is fine as the first step for us to get it running on 8.0. I would be more concerned if we planned on doing a bunch of SQLA work in the near future and had mismatched versions. We should just prioritize getting us to 8.0 everywhere.

PR #306, opened 2022-01-07
Move to Focal Fossa

squeaky-pl:
This upgrades the base image from Ubuntu 18.04 (Bionic Beaver) to current Ubuntu 20.04.3 LTS (Focal Fossa).

The default Python version on this image is Python 3.8.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:bionic-20210930
+FROM ubuntu:focal-20220105
 ARG PYTHON_VERSION=3.8
 
 RUN groupadd -g 5000 sync-engine \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -34,7 +34,7 @@ RUN DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get dist-upgrade -y
   software-properties-common \
   && rm -rf /var/lib/apt/lists/*
 
-RUN if [ "${PYTHON_VERSION}" != "3.6" ] ; \
```
jpmelos, line 37: A couple of dumb questions about other lines in this file that I don't fully understand or remember what they do:

1. In line 42, do we still need the `= 3.8` condition?
2. In line 45, does that `2.7` refer to the Python version, and does it still make sense? Should it change?

squeaky-pl: Line 42 still applies, debian distros for Python 3.8 and newer split `python` package and into `python` and `python-distutils`. We need `python-distutils` to build some packages from source.

In line 45 we indeed don't need 2.7 anymore, it meant last version of pip that works on 2.7 and we don't care about 2.7 anymore. I changed that.

jpmelos (approved):

PR #307, opened 2022-01-07
Simplify running tests

squeaky-pl:
You can now run tests without specifying tests/ manually and also you do not need to pass env variable.

`docker-compose run app pytest` should be enough.


thomasst (approved):
Looks good (I ran it with the m1 branch merged). Can we also add a mention to the README on how to run the tests?

```
========== 626 passed, 15 skipped, 1036 warnings in 212.26s (0:03:32) ==========
```

PR #308, opened 2022-01-13
Attempt adding foreign keys back to messagecategory

squeaky-pl:


jkemp101:
I used `set foreign_key_checks=OFF` for this change https://github.com/closeio/nylas/pull/63#issuecomment-532651739 and I think it worked fine.

PR #310, opened 2022-01-18
replaced .format() with fstrings in console.py

katherinekemp:
Simplified .format() calls to fstrings in console.py now that program runs in Python 3.


squeaky-pl (approved):
LGTM. Thank you.

PR #314, opened 2022-02-14
Address remaining flake8-simplify lints that could not be done on Python 2

squeaky-pl:
Partly addresses: https://github.com/closeio/sync-engine/issues/311

PR #315, opened 2022-02-14
Tweaks to how we document exceptions in suppress

jpmelos:
(Note that this is merging into https://github.com/closeio/sync-engine/pull/314.)

---

While reviewing [the other PR](https://github.com/closeio/sync-engine/pull/314), I thought that maybe the comments in some `try...except` clauses lost an important part of their meaning in translation: they look like they apply to the code in question unconditionally, instead of only when an exception is raised.

What I propose here is a notation to make it clear that that comment only applies if an exception covered by the `suppress` manager is raised:

```python
with suppress(AnException):
    # AnException: A comment about when the exception is raised.
    the_code_here()
```

What do you think?

### DIFF
--- a/inbox/actions/backends/generic.py
+++ b/inbox/actions/backends/generic.py
@@ -151,8 +151,8 @@ def remote_delete_folder(crispin_client, account_id, category_id):
         display_name = category.display_name
 
     with contextlib.suppress(IMAP4.error):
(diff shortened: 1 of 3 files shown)


squeaky-pl (approved):
Makes sense, I had the same impression about context being lost but was not sure what would be the solution. This looks better.

PR #322, opened 2022-02-26
Emit single bulk insert

squeaky-pl:


inbox/contacts/processing.py
```diff
@@ -128,8 +128,11 @@ def update_contacts_from_event(db_session, event, namespace_id):
             return
 
         contact_map = _get_contact_map(db_session, namespace_id, all_addresses)
+        db_session.add_all(contact_map.values())
+        db_session.flush()
```
squeaky-pl, line 132: add and flush here are needed so we know `contact_id` later

PR #327, opened 2022-03-09
FORCE INDEX

squeaky-pl:
Fixes #

Summary:

Test Plan:

Reviewers:


wojcikstefan:
@squeaky-pl is this something that we need still? If not, can probably close the PR

PR #328, opened 2022-04-28
Switch to always up-to-date focal (20.04) image for quicker builds.

dsegan:


Dockerfile
```diff
@@ -1,46 +1,55 @@
...
   then \
     add-apt-repository ppa:deadsnakes/ppa; \
-  fi; \
-  DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y python"${PYTHON_VERSION}"-dev; \
-  if [ "${PYTHON_VERSION}" = "3.8" ] || [ "${PYTHON_VERSION}" = "3.9" ] ; then DEBIAN_FRONTEND=noninteractive apt-get install -y python"${PYTHON_VERSION}"-distutils; fi; \
-  rm -rf /var/lib/apt/lists/*
+  fi \
+  && apt-get update \
+  && apt-get install -y python"${PYTHON_VERSION}"-dev \
+  && if [ "${PYTHON_VERSION}" = "3.7" ] \
```
dsegan, line 45: 3.7 also requires python3.7-distutils to work.

dsegan, line 45: 3.7 also requires python3.7-distutils to work.

AlecRosenbaum: can we remove the if condition then? What do we support more than 3.7 / 3.8 / 3.9?

dsegan: docker-compose.yml devtools service [references circle/python:3.6 image](https://github.com/closeio/sync-engine/blob/master/docker-compose.yml#L42), though if that works with 3.7, we can just move that up and get rid of the `if` indeed.

AlecRosenbaum: Yep, we should do that. No need to support 3.6

dsegan: Done in f99292a83a7cc04eac96b515c5f2d5a158280b2d.

AlecRosenbaum (approved):

PR #329, opened 2022-05-09
Refactor docker image to a multi-stage image for smaller final image size.

dsegan:
Introduces 3 stages:
 - `base_python_image` which sets up the sync_engine user and installs appropriate PYTHON_VERSION version of Python.
 - `dependency_image` which builds the full virtualenv for the final image
 - final production image that includes only runtime dependencies (likely some still missing)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,39 @@
-FROM ubuntu:20.04
+FROM ubuntu:20.04 AS base_python_image
 ARG PYTHON_VERSION=3.8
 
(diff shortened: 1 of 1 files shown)


dsegan:
For this public repository, I think it's nice that we support multiple Python versions: external users can choose any one of the supported versions.  Though I am not sure if we've got any external users :)

AlecRosenbaum:
If it just runs as a docker service accessible via HTTP/HTTPS then there's no reason someone should want to run using version besides what we support/test, right?

PR #335, opened 2022-07-20
Update pip

vtclose:
More recent pip version and cleaning up dependency installation commands.


squeaky-pl:
You would wonder why sync-engine uses pymongo. They use pymongo way of serializing JSON, I was too scared to find out why... maybe it would break some data we have in production.

vtclose:
I pinned setuptools to the last version that still works with same pymongo. Someone else will have to figure out the pymongo situation later: either upgrade it or vendor/backport json utils from there (although licenses are not super compatible for that).

squeaky-pl (approved):

PR #337, opened 2022-07-21
Only build with system-provided python 3.8

vtclose:
This will reduce maintenance effort required for keeping dependencies up to date and alert-free.

Note that new images are build without `-py3.x` suffix.


squeaky-pl (approved):
LGTM

PR #349, opened 2022-09-15
Merge HTTPFrontend with ProfilingHTTPFrontend

squeaky-pl:
The HTTPFrontend base class is only inherited by ProfilingHTTPFrontend which means it should not be a separate entity. This also confuses mypy.


wojcikstefan (approved):
Really love this type of cleanup! ⭐

PR #350, opened 2022-09-16
Mypy: module by module

squeaky-pl:
Step by step. This adds annotations module by module, as turning mypy on the whole codebase would be hard and a big PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ jobs:
 
       - run:
           name: mypy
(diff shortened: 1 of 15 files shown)


drewler (approved):
👍 

New modules should be added [here](https://github.com/closeio/sync-engine/pull/350/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R34) is that correct?

PR #352, opened 2022-09-19
Upgrade Mako to 1.2.2

jpmelos:
This is to clear a security alert:

191034016-79025d1e-73ac-4331-8b7b-0ac827f93714

Mako's changes:

https://github.com/sqlalchemy/mako/releases

I don't see anything concerning. The most impactful changes are related to Python 2 to 3 still, and dropping Python 3.6 support.

Tests pass locally.


squeaky-pl (approved):
I believe we only use mako to generate alembic migrations in this project so this is safe.

Issue #338, opened 2022-08-17
Include required scopes when requesting an access token to AuthAlligator


drewler:
Once `authalligator` and `authalligator-client` have been upgraded to support requesting specific scopes, we should update the places where it's called to add the required scopes.

Main place seems to be [`_new_access_token_from_authalligator`](https://github.com/closeio/sync-engine/blob/e19ff46f3ca8a1cb7fcfbdb6d0cc3cd10fe1a063/inbox/auth/oauth.py#L79-L129). The `OAuthAuthHandler ` subclasses seems to have that info already (`OAUTH_SCOPE`) Eg:

https://github.com/closeio/sync-engine/blob/e19ff46f3ca8a1cb7fcfbdb6d0cc3cd10fe1a063/inbox/auth/microsoft.py#L39-L49

drewler:
Since:
- We want to get rid of `verify_account`
- It doesn't really help with [what the code suggest it's trying to do](https://github.com/closeio/sync-engine/blob/e19ff46f3ca8a1cb7fcfbdb6d0cc3cd10fe1a063/inbox/auth/oauth.py#L97-L98) (it won't really "force a refresh", it will just try to check the token with an endpoint on the third party service)
- The new scoped tokens that will be generated can't be verified against that Graph endpoint anyways

I think it will be a good chance to remove the call to `verify_account`.

PR #353, opened 2022-09-26
Update `authalligator-client`

drewler:
Implements #338

- Update usage of `authalligator-client` (pending merge of https://github.com/closeio/authalligator-client/pull/12).
  - Removes usage of `verify` mutation since it wasn't really causing a forced refresh as the code was expecting. The mutation will be removed soon and a new `refresh` parameter will be added that truly allows forcing a token refresh.
- Add type annotations for the affected code path
- Define scope properties on `OAuthAccount` subclasses (eg: [Outlook accounts](https://github.com/closeio/sync-engine/blob/0ae9022ed7fd88d8dfc75a28e75c45a4a8027329/inbox/models/backends/outlook.py#L36-L38)).
  - Those are forwarded from an initial [`get_token`](https://github.com/closeio/sync-engine/blob/57426ee16cef81ebf0b2fe5ff946a04022551b25/inbox/auth/oauth.py#L176-L178) call until they reach [`_new_access_token_from_authalligator`](https://github.com/closeio/sync-engine/blob/57426ee16cef81ebf0b2fe5ff946a04022551b25/inbox/auth/oauth.py#L165-L171) (they are passed only when the tokens are requested to `authalligator`).

### DIFF
--- a/inbox/auth/microsoft.py
+++ b/inbox/auth/microsoft.py
@@ -36,7 +36,7 @@ class MicrosoftAuthHandler(OAuthAuthHandler):
     )
     OAUTH_USER_INFO_URL = "https://outlook.office.com/api/v2.0/me"
 
-    OAUTH_SCOPE = " ".join(
+    OAUTH_AUTH_SCOPES = " ".join(
         [
             "https://outlook.office.com/IMAP.AccessAsUser.All",
             "https://outlook.office.com/SMTP.Send",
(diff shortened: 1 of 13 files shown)


inbox/auth/microsoft.py
```diff
@@ -36,7 +36,7 @@ class MicrosoftAuthHandler(OAuthAuthHandler):
     )
     OAUTH_USER_INFO_URL = "https://outlook.office.com/api/v2.0/me"
 
-    OAUTH_SCOPE = " ".join(
+    OAUTH_AUTH_SCOPES = " ".join(
```
cschams, line 39: nit: If it can be consistent, I would prefer to use "scope" := string, and "scopes" := list. That works at least with the OAUTH spec and our AuthAlligator API. Consistent naming of these constants isn't critical though, since editors and linting will identify them correctly.

cschams (approved):
Looks good to me!

PR #354, opened 2022-09-26
Mypy: module by module 2

squeaky-pl:


inbox/events/ical.py
```diff
@@ -709,7 +709,7 @@ def send_rsvp(ical_data, event, body_text, status, account):
     rsvp_to = rsvp_recipient(event)
 
     if rsvp_to is None:
-        raise SendMailException("Couldn't find an organizer to RSVP to.")
+        raise SendMailException("Couldn't find an organizer to RSVP to.", None)
```
squeaky-pl, line 712: This needs a http_code but it does not make sense to pass one here.

inbox/sendmail/smtp/postel.py
```diff
@@ -205,9 +204,9 @@ def _smtp_oauth2_try_refresh(self):
     def _try_xoauth2(self):
         auth_string = "user={}\1auth=Bearer {}\1\1".format(
             self.email_address, self.auth_token
-        )
+        ).encode()
```
squeaky-pl, line 207: We don't do SMTP sending from sync-engine, but this was invalid because b64encode needs bytes and returns bytes

drewler (approved):

PR #358, opened 2022-10-03
More mypy

squeaky-pl:


inbox/crispin.py
```diff
@@ -916,7 +918,7 @@ def find_by_header(self, header_name, header_value):
         matching_draft_headers = self.fetch_headers(all_uids)
         results = []
         for uid, response in matching_draft_headers.items():
-            headers = response["BODY[HEADER]"]
+            headers = response[b"BODY[HEADER]"]
```
squeaky-pl, line 921: the IMAP responses are always in bytes, this function is only called from logic that deletes emails, and we don't do it, so we did not catch this earlier.

drewler (approved):

PR #359, opened 2022-10-04
Filter out empty folder names

squeaky-pl:
As discovered on one Outlook account you can get `None` as IMAP folder name, we cannot handle those so just skip them from syncing. The Outlook account in question has valid INBOX folder apart from malformed folder.

### DIFF
--- a/inbox/crispin.py
+++ b/inbox/crispin.py
@@ -376,7 +376,14 @@ class CrispinClient:
         IMAPClient parses this response into a list of
         (flags, delimiter, name) tuples.
         """
-        return self.conn.list_folders()
+
+        # As discovered in the wild list_folders() can return None as name,
+        # we cannot handle those folders anyway so just filter them out.
+        return [
+            (flags, delimiter, name)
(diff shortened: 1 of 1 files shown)


inbox/crispin.py
```diff
@@ -376,7 +376,14 @@ def _fetch_folder_list(self):
...
         (flags, delimiter, name) tuples.
         """
-        return self.conn.list_folders()
+
+        # As discovered in the wild list_folders() can return None as name,
+        # we cannot handle those folders anyway so just filter them out.
+        return [
+            (flags, delimiter, name)
+            for flags, delimiter, name in self.conn.list_folders()
+            if name
```
drewler, line 385: Is a folder valid when it has an empty string as its name? Maybe we should check `if name is None` instead.

I agree about skipping invalid folders 👍

drewler (approved):

inbox/crispin.py
squeaky-pl: I've actually searched our production data and there are no empty string folder names, but I don't mind checking `None` explicitely.

PR #363, opened 2022-10-06
More mypy fixes

squeaky-pl:


inbox/heartbeat/store.py
```diff
@@ -115,9 +116,10 @@ def publish(self, key, timestamp):
         # Update indexes
         self.update_folder_index(key, float(timestamp))
 
-    def remove(self, key, device_id=None):
+    def remove(self, key, device_id=None, client=None):
```
squeaky-pl, line 119: This is fun.

The [function that calls this with third parameter](https://github.com/closeio/sync-engine/blob/415563a560e8800dba829dc43310c9d38f6a4add/inbox/heartbeat/store.py#L134-L157) is decorated by a decorator that eats up exception. In case it calls it with third parameter it already creates a pipeline to make redis calls in batch. Since this eats up exception we never noticed that.

PR #367, opened 2022-10-08
Support events without dtend but with duration in ICS

squeaky-pl:
We have a related Rollbar.

The code that parses ICS files to create events only admitted events that had **dtstart** and **dtend** fields. That's not the only way to specify event times, you can have **dtstart** and **duration** as well and currently we needlessly choke with Rollbar on such ICS files.


inbox/events/ical.py
```diff
@@ -58,9 +58,14 @@ def events_from_ics(namespace, calendar, ics_str):
...
             try:
                 original_start = component.get("dtstart").dt
-                original_end = component.get("dtend").dt
+                if component.get("dtend"):
+                    original_end = component.get("dtend").dt
+                else:
+                    original_end = original_start + component.get("duration").dt
             except AttributeError:
-                raise MalformedEventError("Event lacks start and/or end time")
+                raise MalformedEventError(
```
squeaky-pl, line 66: I will eventually turn this to warning so we don't get Rollbars, but first I want to try to recover from all the cases we can recover from because there are more combinations allowed.

drewler (approved):

PR #380, opened 2022-10-19
Refactor When

squeaky-pl:
This removes badly used `@abc.abstractproperty` that was used assigned on class properties instead of decorating instance properties. Mypy complains about it and also newer flake-bugbear will also say no.

### DIFF
--- a/inbox/models/when.py
+++ b/inbox/models/when.py
@@ -1,6 +1,5 @@
-import abc
 import contextlib
-from typing import Union
+from typing import List, Type, Union
 
 import arrow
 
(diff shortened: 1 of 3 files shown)


inbox/models/when.py
```diff
@@ -65,40 +59,36 @@ def parse_keys(cls, raw):
...
         return self.end - self.start
 
-    def get_time_dict(self):
+    def get_time_dict(self) -> Dict[str, arrow.Arrow]:
         times = (self.start, self.end)
         return dict(zip(self.json_keys, times))
 
 
-class AllDayWhen(When):
-    all_day = True
```
squeaky-pl, line 93: This class was used just to set `all_day = True`, which is too little to have a separate class

drewler (approved):

PR #395, opened 2022-10-31
Microsoft: parse_event, parse_calendar

squeaky-pl:
Provides functions that convert Microsoft Graph Responses to sync-engine ORM objects.

### DIFF
--- a/inbox/events/google.py
+++ b/inbox/events/google.py
@@ -1,5 +1,6 @@
 """Provide Google Calendar events."""
 import datetime
+import email.utils
 import json
 import random
 import time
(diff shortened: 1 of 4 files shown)


inbox/events/google.py
```diff
@@ -536,8 +537,8 @@ def parse_event_response(event: Dict[str, Any], read_only_calendar: bool) -> Eve
     creator = event.get("creator")
 
     if creator:
-        owner = "{} <{}>".format(
```
squeaky-pl, line 539: This would not handle properly names containing characters like `<` i.e. pair `"Example <"`, `"test@example.com"` should be encoded as `'"Example <" <test@example.com>'` and not `"Example < <test@example.com>"`

cschams (approved):

PR #397, opened 2022-11-01
Microsoft: Extract is_default_calendar

squeaky-pl:
We need to know in sync-engine if your Calendar is default/primary or not. This is because when syncing we always start by syncing your default calendar as it's most likely the one you care about.

In case of Google you can know if a calendar is default/primary by comparing it's uid with account email address. In case of Microsoft calendar uids are opaque and Ms Graph API has dedicated default field that carries this information.

### DIFF
--- a/inbox/api/kellogs.py
+++ b/inbox/api/kellogs.py
@@ -20,6 +20,7 @@ from inbox.models import (
     Thread,
     When,
 )
+from inbox.models.calendar import is_default_calendar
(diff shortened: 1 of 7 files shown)


squeaky-pl:
SQL migration to run **before** deploy

```sql
ALTER TABLE inbox.calendar ADD `default` tinyint(1) NULL;
UPDATE inbox.alembic_version SET version_num = '7bb5c0ca93de' WHERE version_num = '52783469ee6c';
```

cschams (approved):

PR #401, opened 2022-11-04
Microsoft: Extract AbstractEventsProvider

squeaky-pl:
This extract abstract events provider from GoogleEventsProvider. This will be later extended to implement MicrosofEventsProvider.

### DIFF
--- a/inbox/events/google.py
+++ b/inbox/events/google.py
@@ -14,19 +14,16 @@ import requests
 from inbox.auth.oauth import OAuthRequestsWrapper
 from inbox.basicauth import AccessNotEnabledError, OAuthError
 from inbox.config import config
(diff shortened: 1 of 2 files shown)


inbox/events/google.py
```diff
@@ -206,25 +183,25 @@ def _get_resource_list(self, url: str, **params) -> List[Dict[str, Any]]:
                     token = self._get_access_token(force_refresh=True)
                     continue
                 elif r.status_code in (500, 503):
-                    log.warning("Backend error in calendar API; retrying")
+                    self.log.warning("Backend error in calendar API; retrying")
```
squeaky-pl, line 186: Notice that this was logging using global logger before, so the account_id info was missing on the logger

cschams (approved):

PR #418, opened 2022-11-15
Fix event ownership transfer for Google Calendar

drewler:
We extracted event ownership (`Event.owner`) info for Google Calendar events from `creator`, which doesn't change after transferring ownership. Oddly, we were computing `is_owner` from `organizer`, which is correct.

This PR uses `organizer` data to compute `owner`. In theory, it shouldn't be necessary to check `creator` (old behavior), but I left that check in case there's no `organizer` data (for some unknown reason).

https://github.com/closeio/sync-engine/blob/cdfdf6ab363cd2b17af9d5b0f745615c8fdc1f2a/inbox/events/google.py#L528-L536

### DIFF
--- a/inbox/events/google.py
+++ b/inbox/events/google.py
@@ -522,13 +522,18 @@ def parse_event_response(event: Dict[str, Any], read_only_calendar: bool) -> Eve
 
     # Ownership, read_only information
     creator = event.get("creator")
+    organizer = event.get("organizer")
+    is_owner = bool(organizer and organizer.get("self"))
 
-    if creator:
+    owner = ""
+    if organizer:
+        owner = email.utils.formataddr(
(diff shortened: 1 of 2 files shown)


squeaky-pl (approved):
Not sure how Google APIs work but it makes sense :-)

drewler:
> Not sure how Google APIs work but it makes sense :-)

Should have linked the [API docs](https://developers.google.com/calendar/api/v3/reference/events). I also tested manually:
- Created an event with "User A" and "User B"
  - "User A" being the creator and original owner/organizer
- Transferred ownership of the event to "User B"
- Checked the `Event` attributes' data on `sync-engine`, "User A" was still the "owner"
- Checked the `raw_data` for the `Event` on `sync-engine`. `organizer` was set to "User B", so after checking the API docs I decided to use that.

PR #420, opened 2022-11-16
Microsoft: Events provider MVP

squeaky-pl:
This provides Minimum Viable Product™️ version of Microsoft Calendar Sync.

Currently supports:
  * Singular events
  * Recurring events (but without gaps and edits inside series yet)
  
Currently does not support:
   * Webhook subscriptions (but it will resync every 5 minutes so changes eventually get there but just with a lot delay)
   * Deletions
   
Tested this locally with CRM and things seem to be working just fine. More PRs to provide missing parts will follow.

### DIFF
--- a/inbox/events/microsoft/parse.py
+++ b/inbox/events/microsoft/parse.py
@@ -619,7 +619,7 @@ def parse_event(
         all_day=all_day,
         owner=owner,
         is_owner=is_owner,
-        ready_only=read_only,
+        read_only=read_only,
         participants=participants,
         recurrence=recurrence,
         last_modified=last_modified,
--- a/inbox/models/backends/oauth.py
+++ b/inbox/models/backends/oauth.py
@@ -24,7 +24,9 @@ def hash_token(token, prefix=None):
     return sha256(string.encode()).hexdigest()
 
 
-def log_token_usage(reason, refresh_token=None, access_token=None, account=None):
+def log_token_usage(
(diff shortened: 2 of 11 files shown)


inbox/models/backends/oauth.py
```diff
@@ -34,41 +36,65 @@ def log_token_usage(reason, refresh_token=None, access_token=None, account=None)
...
 
     def get_token(
         self,
         account: "OAuthAccount",
         force_refresh: bool = False,
         scopes: Optional[List[str]] = None,
     ) -> str:
-        if account.id in self._tokens:
-            token, expiration = self._tokens[account.id]
+        cache_key = self.get_cache_key(account, scopes)
```
squeaky-pl, line 55: Those changes here are needed so we don't get tokens with wrong scopes from the cache. I was getting "Invalid audience" errors from Microsoft Graph and I realized this is because of this local in-process cache. Instead of caching on account ids we cache on pair (account_id, scopes) which prevents those errors.

inbox/events/microsoft/parse.py
```diff
@@ -619,7 +619,7 @@ def parse_event(
         all_day=all_day,
         owner=owner,
         is_owner=is_owner,
-        ready_only=read_only,
```
squeaky-pl, line 622: Oopsie

cschams (approved):
Looks good to me

PR #426, opened 2022-11-28
Microsoft: Calendar deletes

squeaky-pl:
This is implemented differently than in Google again. You cannot fetch deleted calendars with Microsoft (like we do in Google). We fetch them all anyway in each synchronization iteration and people typically have only a couple of calendars so one can just compare what comes from the API with what we have in the database.


cschams (approved):
Looks good to me. I agree, calendar lists should be fairly short.

PR #429, opened 2022-11-29
Microsoft: Event organizer can be None

squeaky-pl:
While testing with real data I found that Microsoft Graph can return events with `"organizer"` set to `None`. We are rollbaring currently for those events.

### DIFF
--- a/inbox/events/microsoft/graph_types.py
+++ b/inbox/events/microsoft/graph_types.py
@@ -194,7 +194,7 @@ class MsGraphEvent(TypedDict):
     end: MsGraphDateTimeTimeZone
     lastModifiedDateTime: str
     showAs: MsGraphShowAs
-    organizer: MsGraphRecipient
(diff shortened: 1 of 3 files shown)


drewler (approved):
Not really related to this bug, but I realized that for Microsoft events maybe we should do another attempt to guess the `owner` when there's no `organizer` data? Eg: for Google events we try to extract that from `creator` (not sure what's the equivalent for Microsoft)

https://github.com/closeio/sync-engine/blob/cdfdf6ab363cd2b17af9d5b0f745615c8fdc1f2a/inbox/events/google.py#L533-L536

Separate question: Do we know if there's a common pattern for events that don't have `organizer`? Maybe they are events generated outside Outlook/Microsoft?

squeaky-pl:
@drewler I am not aware of any fallback: this is the list of properties events have: https://learn.microsoft.com/en-us/graph/api/resources/event?view=graph-rest-1.0#properties

PR #444, opened 2022-12-12
Demote expected to create imapuid

squeaky-pl:
As noted [here ](https://github.com/closeio/sync-engine/blob/3fc0382512bba63e1a9fe551910c2a069b20615a/inbox/mailsync/backends/imap/generic.py#L540-L541) the imapuid can be already created due to race conditions, the logic already does the correct thing  in such case and there's no need to have an error here.

### DIFF
--- a/inbox/mailsync/backends/imap/generic.py
+++ b/inbox/mailsync/backends/imap/generic.py
@@ -559,7 +559,7 @@ class FolderSyncEngine(Greenlet):
             return None
 
         # Check if the message is valid.
-        if raw_message.body:
+        if not raw_message.body:
             log.warning("Server returned a message with an empty body.")
             return None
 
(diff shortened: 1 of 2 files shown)


inbox/mailsync/backends/imap/generic.py
```diff
@@ -534,35 +534,37 @@ def poll_for_changes(self):
...
                 "Expected to create imapuid, but existing row found",
-                remote_msg_uid=msg.uid,
+                remote_msg_uid=raw_message.uid,
                 existing_imapuid=existing_imapuid.id,
             )
             return None
 
         # Check if the message is valid.
-        if msg.body is None:
+        if not raw_message.body:
```
drewler, line 562: I guess those two will be always equivalent? Can a message be valid when `raw_message.body` is `""`?

squeaky-pl: it can be actually `""` but never `None`. As soon as put typing info mypy catched this because this code was unreachable and we should never process empty messages.

PR #445, opened 2022-12-12
Bump pip and virtualenv

squeaky-pl:


Dockerfile
```diff
@@ -17,6 +17,7 @@ RUN echo $BUILD_WEEK && apt-get update \
     g++ \
     git \
     python3-dev \
+    python3-pip \
```
squeaky-pl, line 20: We used to use `https://bootstrap.pypa.io/pip/get-pip.py` to boostrap pip because it worked on Python 2 as well, now that we are on 3 only we can use distro one and then update it using pip itself.

drewler (approved):

PR #448, opened 2022-12-14
Drop error sampling

squeaky-pl:
This drops the logic we used to sample some noisy Rollbars.

### DIFF
--- a/inbox/error_handling.py
+++ b/inbox/error_handling.py
@@ -1,9 +1,6 @@
-import functools
 import json
 import logging
 import os
-import random
-import re
 import sys
 
 import rollbar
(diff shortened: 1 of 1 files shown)


inbox/error_handling.py
```diff
@@ -75,9 +72,6 @@ def log_uncaught_errors(logger=None, **kwargs):
 
 
 GROUP_EXCEPTION_CLASSES = [
-    "GreenletExit",
-    "LoopExit",
-    "ResourceClosedError",
```
squeaky-pl, line 80: We no longer get those.

drewler (approved):

PR #453, opened 2022-12-19
Microsoft: Probe if subscriptions work in webhook_notifications_enabled

squeaky-pl:
This is currently rollbaring for one account in production.

Sync-engine has two modes for syncing calendar:
* polling changes since last time it synced something, this has more delay of course 
* listening to webhooks, almost instant
	
Depending on what you return from ` webhook_notifications_enabled` it will do one of those two things.

For Microsoft I expected subscriptions to always work, but it turns out that it does not always work in practice. We can still fetch events and calendars but not create subscriptions in this case.

As explained in the docstring:

We found that in practice subscriptions don't work for some accounts. There are some theories on the internet why it does not work i.e.: Office365 administrator applying a restrictive policy or some weird setup when the calendars might still be on on-premise servers but everything else in Azure. Microsoft does not give a definite answer, so we can only speculate.

For more context see:
* https://learn.microsoft.com/en-us/answers/questions/417261/error-on-adding-subscription-on-events-using-ms-gr.html
* https://stackoverflow.com/questions/65030751/ms-graph-adding-subscription-returns-extensionerror-and-serviceunavailable

The good news is that we can fallback to polling mode and still sync those calendars/events. Just slower and with more overhead.

### DIFF
--- a/inbox/events/remote_sync.py
+++ b/inbox/events/remote_sync.py
@@ -49,7 +49,9 @@ class EventSync(BaseSyncMonitor):
     ):
         bind_context(self, "eventsync", account_id)
         self.provider = provider_class(account_id, namespace_id)
-        self.log = logger.new(account_id=account_id, component="calendar sync")
+        self.log = logger.new(
+            account_id=account_id, component="calendar sync", provider=provider_name
+        )
 
         BaseSyncMonitor.__init__(
             self,
@@ -294,6 +296,7 @@ class WebhookEventSync(EventSync):
             account = db_session.query(Account).get(self.account_id)
 
             if not self.provider.webhook_notifications_enabled(account):
+                self.log.warning("Webhook notifications disabled")
                 return
 
             if account.needs_new_calendar_list_watch():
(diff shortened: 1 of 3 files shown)


inbox/events/remote_sync.py
```diff
@@ -49,7 +49,9 @@ def __init__(
     ):
         bind_context(self, "eventsync", account_id)
         self.provider = provider_class(account_id, namespace_id)
-        self.log = logger.new(account_id=account_id, component="calendar sync")
+        self.log = logger.new(
```
squeaky-pl, line 52: This is just a small tweak so I can filter logs by calendar_sync for microsoft (and exclude Google). Mail sync greenlets already have provider set.

drewler (approved):
Do we currently expose through sync-engine's API if a "running" account is using subscriptions or polling? If not might be useful to add it at some point, answer questions like "this account says it's syncing but events are taking too long to show up, why does that happen?" Just thought about it, but I think it can probably wait until it actually becomes a problem/complaint.

squeaky-pl:
@drewler we actually don't, we also don't save this info anywhere. We just call this function each time a greenlet starts which is when you add a new account or restart sync pods. Arguably the situation is now better because you can query logs for the warning I added.

PR #455, opened 2022-12-19
Switch CircleCI runner to 22.04

squeaky-pl:
This is in preparation for Ubuntu 22.04. Note that this only changes the CircleCI runner image, and not Dockerfile for now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,7 +101,7 @@ jobs:
         default: true
         description: Set to false to disable pushing image
     machine:
-      image: ubuntu-2004:202101-01
+      image: ubuntu-2204:current
     resource_class: <<parameters.resource_class>>
     working_directory: ~/sync-engine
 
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -26,6 +26,7 @@ services:
        - MINIO_DOMAIN=minio
 
   app:
+    image: sync-engine_app
```
squeaky-pl, line 29: Docker-compose bundled in newer image by default names images `sync-engine-app` (underscore versus hyphen). We want the old behavior so everything else keeps working.

.circleci/config.yml
```diff
@@ -101,7 +101,7 @@ jobs:
         default: true
         description: Set to false to disable pushing image
     machine:
-      image: ubuntu-2004:202101-01
+      image: ubuntu-2204:current
```
drewler, line 104: Shouldn't we pin a specific one?

drewler (approved):

PR #458, opened 2022-12-22
Microsoft: handle sporadic underlying connection was closed, cache webhook_notifications_enabled

squeaky-pl:
This takes care of two things:

- When I changed `webhook_notifications_enabled` to create a dummy subscription and then delete it I did not realize this is called periodically in a loop. We need to do it only once to know if we are on unlucky account, then we can just cache it.

- We sporadically get `The underlying connection was closed: A connection that was expected to be kept alive was closed by the server.`. We get this when subscribing to notifications over webhooks Microsoft makes an initial contact with webhook url. This is safe to retry.

### DIFF
--- a/inbox/events/microsoft/events_provider.py
+++ b/inbox/events/microsoft/events_provider.py
@@ -73,6 +73,7 @@ class MicrosoftEventsProvider(AbstractEventsProvider):
         self.client = MicrosoftGraphClient(
             lambda: self._get_access_token(scopes=MICROSOFT_CALENDAR_SCOPES)
         )
+        self._webhook_notifications_enabled: Optional[bool] = None
 
     def sync_calendars(self) -> CalendarSyncResponse:
         """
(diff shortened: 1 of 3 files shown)


inbox/events/microsoft/events_provider.py
```diff
@@ -216,6 +217,12 @@ def webhook_notifications_enabled(self, account: Account) -> bool:
         * https://learn.microsoft.com/en-us/answers/questions/417261/error-on-adding-subscription-on-events-using-ms-gr.html
         * https://stackoverflow.com/questions/65030751/ms-graph-adding-subscription-returns-extensionerror-and-serviceunavailable
         """
+
+        # First check if we already have cached value since this function is called
+        # repeatedly and there is no need to do extra HTTP request every time.
+        if self._webhook_notifications_enabled is not None:
+            return self._webhook_notifications_enabled
```
wojcikstefan, line 224: Is it possible that this method is called one time for Account A and another time for Account B? If so, this will have issues. If not, should this method not accept the `account` argument?

squeaky-pl: No. It's always the same account object, I know it's weird given you receive account as parameter but that's how it was done in the past for some reason.

It's:

- single greenlet that is responsible for account sync 
- that has single events provider for this specific account
- that has single client for this specific account

squeaky-pl: I guess this was done like that with Google because they did not want to make repetitive ORM calls in this component, see [Google one](https://github.com/closeio/sync-engine/blob/b606e71f04625e591ffe5220e80fae38d8dda9de/inbox/events/google.py#L287-L409).

squeaky-pl: But I guess the whole thing could be refactored to accept `Account` instance in the constructor.

wojcikstefan: Yea, that would make for sense, I think. Should probably note it as an issue or maybe at least leave a TODO comment.

squeaky-pl: See: https://github.com/closeio/sync-engine/pull/459

PR #463, opened 2023-01-03
remove pinned setuptools

nsaje:
https://github.com/pypa/setuptools/issues/2352 which was the reason for pinning was resolved (and we also upgraded to Python 3.8 in the mean time)


squeaky-pl (approved):
Note that you'll need to update the hash in the companion downstream project to get this into final Docker image.

nsaje:
thanks for the reminder!

PR #474, opened 2023-03-08
Handle MSGraph timezone-aware datetimes

neob91-close:
Fixes #

Summary:
MSGraph can return values such as `{"dateTime": "0001-01-01T00:00:00.0000000Z", "timeZone": "tzone://Microsoft/Utc"}` for some calendar events' `start` or `end` fields. This causes `tzinfo.localize(dt)` in `parse_msgraph_datetime_tz_as_utc` to fail, as `tzinfo` is already set.

Test Plan:

Reviewers:

### DIFF
--- a/inbox/events/microsoft/parse.py
+++ b/inbox/events/microsoft/parse.py
@@ -81,6 +81,14 @@ def parse_msgraph_datetime_tz_as_utc(datetime_tz: MsGraphDateTimeTimeZone):
     # so we need to use ciso8601 because Python stdlib only supports up to 6.
     dt = ciso8601.parse_datetime(datetime_tz["dateTime"])
 
+    # Microsoft sometimes returns payloads such as this one:
+    # {"dateTime": "0001-01-01T00:00:00.0000000Z", "timeZone": "tzone://Microsoft/Utc"}
+    # We're removing the timezone from the parsed datetime, otherwise
+    # `tzinfo.localize` would fail. This datetime is obviously not correct
+    # anyway.
+    if dt.tzinfo and dt == datetime.datetime(1, 1, 1, tzinfo=pytz.UTC):
+        dt = dt.replace(tzinfo=None)
+
     return tzinfo.localize(dt).astimezone(pytz.UTC)
 
 
--- a/tests/events/microsoft/test_parse.py
+++ b/tests/events/microsoft/test_parse.py
@@ -51,6 +51,13 @@ def test_get_microsoft_timezone(windows_tz_id, olson_tz_id):
             },
             datetime.datetime(2022, 9, 22, 16, 30, tzinfo=pytz.UTC),
         ),
+        (
+            {
+                "dateTime": "0001-01-01T00:00:00.0000000Z",
+                "timeZone": "tzone://Microsoft/Utc",
+            },
+            datetime.datetime(1, 1, 1, tzinfo=pytz.UTC),
+        ),
     ],
 )
 def test_parse_msggraph_datetime_tz_as_utc(datetime_tz, dt):


inbox/events/microsoft/parse.py
```diff
@@ -81,6 +81,16 @@ def parse_msgraph_datetime_tz_as_utc(datetime_tz: MsGraphDateTimeTimeZone):
     # so we need to use ciso8601 because Python stdlib only supports up to 6.
     dt = ciso8601.parse_datetime(datetime_tz["dateTime"])
 
+    # Microsoft sometimes returns payloads such as this one:
+    # {"dateTime": "0001-01-01T00:00:00.0000000Z", "timeZone": "tzone://Microsoft/Utc"}
+    # Notice the `Z` - the `dateTime` is timezone-aware. This is undocumented
+    # behaviour, but we will handle it here so it doesn't break sync.
+    if dt.tzinfo:
```
squeaky-pl, line 88: should we just instead honor the tzinfo from dateTime always if it's set?

inbox/events/microsoft/parse.py
```diff
@@ -81,6 +81,16 @@ def parse_msgraph_datetime_tz_as_utc(datetime_tz: MsGraphDateTimeTimeZone):
...
+        # We will only humour Microsoft when it comes to UTC. This can be
```
squeaky-pl, line 89:
```suggestion
        # We will only honour Microsoft when it comes to UTC. This can be
```

squeaky-pl (commented):
See comments

squeaky-pl:
Not surprised, I've seen a lot of weird behaviors while developing this. I would prefer we recovered instead by trusting timeZone on dateTime field.

inbox/events/microsoft/parse.py
neob91-close: I wrote `humour` on purpose 😅 We're humouring Microsoft in that we're letting them return inconsistent data, and we're trying to process it without failing.

inbox/events/microsoft/parse.py
neob91-close: We have no evidence of Microsoft returning values other than `0001-01-01T00:00:00.0000000Z`, so I'd like to minimise the changes to the logic that we do here, and only handle the UTC use-case.

I think it's better to fail for something we don't expect here (which we do already anyway), than to store an incorrectly calculated value by accident.

What do you think?

inbox/events/microsoft/parse.py
```diff
@@ -81,6 +81,14 @@ def parse_msgraph_datetime_tz_as_utc(datetime_tz: MsGraphDateTimeTimeZone):
     # so we need to use ciso8601 because Python stdlib only supports up to 6.
     dt = ciso8601.parse_datetime(datetime_tz["dateTime"])
 
+    # Microsoft sometimes returns payloads such as this one:
+    # {"dateTime": "0001-01-01T00:00:00.0000000Z", "timeZone": "tzone://Microsoft/Utc"}
+    # We're removing the timezone from the parsed datetime, otherwise
+    # `tzinfo.localize` would fail. This datetime is obviously not correct
+    # anyway.
+    if dt.tzinfo and dt == datetime.datetime(1, 1, 1, tzinfo=pytz.UTC):
```
squeaky-pl, line 89: Nit, I think it would be enough to
```suggestion
    if dt == datetime.datetime(1, 1, 1, tzinfo=pytz.UTC):
```

squeaky-pl (approved):

inbox/events/microsoft/parse.py
neob91-close: Did it for performance reasons. I'm generally against premature optimisation, but I don't know how often this is called, so I did not want to decrease performance for all executions here, and a single `dt.tzinfo` check doesn't hurt.

By the way, it's `1000ms` vs `70ms` for 1,000,000 calls.

wojcikstefan: For both minor performance gains and more importantly for readability, I suggest we construct `datetime.datetime(1, 1, 1, tzinfo=pytz.UTC)` only once and assign it to a const like `MICROSOFT_NO_DATETIME_SENTINEL_VALUE`, along with a explaining what this is about.

neob91-close: Even if we do this, I'd still like to keep the `dt.tzinfo` check as it's extremely cheap compared to datetime comparison (which could be a bit more expensive when comparing two datetimes using different timezone objects for UTC, e.g. `pytz.UTC` and `datetime.timezone.utc`)

wojcikstefan: Yep makes sense

Issue #302, opened 2021-12-23
Investigate missing foreign keys on `messagecategory` table.


squeaky-pl:
Currently the table looks like this:

```
CREATE TABLE `messagecategory` (
  `created_at` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `message_id` bigint(20) NOT NULL,
  `category_id` bigint(20) NOT NULL,
  PRIMARY KEY (`id`),
  KEY `ix_messagecategory_category_id` (`category_id`),
  KEY `ix_messagecategory_created_at` (`created_at`),
  KEY `message_category_ids` (`message_id`,`category_id`)
);
```

Notice that there are no foreign key constraints on `message_id` and `category_id` columns. This can lead to to data corruption and exceptions on Python side. We should investigate if they were omitted on purpose. If not we should add them.

squeaky-pl:
It looks like they were dropped on purpose https://github.com/closeio/sync-engine/blob/9022236c98ca7abdef67062a38a7a295c498b6e0/migrations/versions/225_drop_messagecategory_foreign_keys.py#L20

squeaky-pl:
The commit that deletes them is this one: https://github.com/closeio/sync-engine/commit/ff8b0bd749ac71f2c8c52430ccb8d3707485a499

PR #475, opened 2023-03-16
Don't break on missing categories

squeaky-pl:
This is a stop gap solution for https://github.com/closeio/sync-engine/issues/302.

There's a pretty laud rollbar both on CRM and sync-engine side.

`category` can be None because somebody at Nylas decided to drop a foreign key mentioned in [the issue](https://github.com/closeio/sync-engine/issues/302). We should still investigate if we can add those constraints back because the decision to drop such constraints does not look like a correct one to me. In the meantime let's ship this because for people that hit this problem we simply get stuck on this and stop syncing their inboxes.

### DIFF
--- a/inbox/models/thread.py
+++ b/inbox/models/thread.py
@@ -117,9 +117,11 @@ class Thread(MailSyncBase, HasPublicID, HasRevisions, UpdatedAtMixin, DeletedAtM
         sorted_messages = sorted(
             self.messages, key=lambda m: m.received_date, reverse=True
         )
-        for m in sorted_messages:
-            if "sent" in [c.name for c in m.categories] or (m.is_draft and m.is_sent):
-                sent_recent_date = m.received_date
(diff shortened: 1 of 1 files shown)


inbox/models/thread.py
```diff
@@ -117,9 +117,11 @@ def most_recent_sent_date(self):
         sorted_messages = sorted(
             self.messages, key=lambda m: m.received_date, reverse=True
         )
-        for m in sorted_messages:
-            if "sent" in [c.name for c in m.categories] or (m.is_draft and m.is_sent):
-                sent_recent_date = m.received_date
+        for message in sorted_messages:
+            if "sent" in [
+                category.name for category in message.categories if category
```
squeaky-pl, line 122: `if category` is the part that I added to filter out `None` categories that come from missing categories because of missing FK.

PR #479, opened 2023-03-29
move initial poll call out of loop

nsaje:
Currently we're [polling the database](https://github.com/closeio/sync-engine/blob/master/inbox/mailsync/service.py#L135) on *every* loop iteration so listening to the Redis queue for events only means we'll start syncing the new account(s) sooner than in 20s which is what the poll interval is set to.

I think the call that I moved was intended to be _outside_ the worker loop so we'd only poll the state of the world on startup and then rely on EventQueue to start syncing new accounts.

WDYT @squeaky-pl  ? I'm not sure if this is worth merging since I don't see this polling to be a significant load on the DB.

### DIFF
--- a/inbox/mailsync/service.py
+++ b/inbox/mailsync/service.py
@@ -124,7 +124,7 @@ class SyncService:
 
     def run(self):
         # When the service first starts we should check the state of the world.
-        self.poll()
+        retry_with_logging(self.poll)
         while self.keep_running:
             retry_with_logging(self._run_impl, self.log)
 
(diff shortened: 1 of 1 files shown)


nsaje:
@squeaky-pl ping

squeaky-pl (commented):
This might be dangerous. If you look at retry_with_logging it catches and retries on several mysql and socket exceptions. poll itself interacts with mysql through `account_ids_to_sync`.

inbox/mailsync/service.py
```diff
@@ -123,6 +123,8 @@ def __init__(
             gevent.spawn_later(exit_after, self.stop)
 
     def run(self):
+        # When the service first starts we should check the state of the world.
+        self.poll()
```
squeaky-pl, line 127:
```suggestion
        retry_with_logging(self.poll)
```

squeaky-pl (approved):

PR #480, opened 2023-04-03
Reduce mail parsing memory footprint, don't do useless processing

squeaky-pl:
TLDR: we had an email message in our production cluster that was causing OOMs while being synced. I downloaded the message, measured memory pressure while parsing it and realized that email prasing logic uses excessive amount of RAM and does pointless operations. 

1. We do pointless `str` -> `bytes` -> `str` conversions while storing html and text parts, we just need `str`s.
2. While calculating message snippet and body we always prefer html parts over plain parts. Most emails these days contain alternative html and plain parts, the plain parts are only carried for the reason of backwards compatibility. Nonetheless we were carrying over plain parts through processing and doing calculations on it even if we were throwing away the results.
3. We were prematurely decoding a mimepart by accessing `mimepart.body` early even if we didn't care about what's inside this particular mimepart.

The message that I saw in production is a pathological case because it's a really long email that contains more than 1000 mimeparts, but it demonstrates very well the kind of savings we can make here.

The video shows memory pressure after processing each mimepart. On master the memory pressure (measurements taken with psutil rss usage) to process this message peaks at 866MBs:

https://user-images.githubusercontent.com/754356/229727005-9a22a436-b97c-4946-a93e-0297636cbd49.mp4

After changes on this branch it's reduced to 190MBs and goes significantly faster.

https://user-images.githubusercontent.com/754356/229727551-c07d8eea-c51b-45f3-a4cf-91ec1a4fbc70.mp4

For obvious privacy reasons I could not ship this email/test, but this logic is already covered by other tests.

### DIFF
--- a/inbox/models/message.py
+++ b/inbox/models/message.py
@@ -489,6 +489,12 @@ class Message(MailSyncBase, HasRevisions, HasPublicID, UpdatedAtMixin, DeletedAt
         if filename == "":
             filename = None
 
+        # Caution: Don't access mimepart.body unless you are sure
+        # you are gonna need it in further processing. Reading this
+        # attribute increases memory pressure siginificantly as it
+        # immediately triggers decoding behind the scenes.
+        # See: https://github.com/closeio/sync-engine/pull/480
+
         is_text = content_type.startswith("text")
         if disposition not in (None, "inline", "attachment"):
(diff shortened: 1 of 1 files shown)


drewler (approved):
🚀

PR #484, opened 2023-04-13
Bump setuptools to 67.4.0 and pymongo to 3.0

neob91-close:
Resolves [CVE-2022-40897](https://github.com/advisories/GHSA-r9hx-vwmv-q579)

We need to bump PyMongo along with setuptools, because bumping setuptools removes 2to3 that PyMongo's installation depends on.

Changelog: https://pymongo.readthedocs.io/en/stable/changelog.html
Since we only use `bson.json_util`, the part that should interest us is this:
https://user-images.githubusercontent.com/113755748/231822387-e2bf782c-4c64-4550-abd0-9a972c58a26a.png

It looks like these changes do not affect us, but I'm not sure about the `Int64` bit.

Also, not 100% sure I understand why we're using the PyMongo bson module.
Is it so that we can store arbitrary data structures (including datetimes) in the database using a string representation of BSON?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,7 +55,7 @@ ENV PATH="/opt/venv/bin:$PATH"
 COPY --chown=sync-engine:sync-engine ./ ./
 RUN python3 -m pip install pip==22.3.1 virtualenv==20.17.1 && \
   python3 -m virtualenv /opt/venv && \
-  /opt/venv/bin/python3 -m pip install setuptools==57.5.0 && \
+  /opt/venv/bin/python3 -m pip install setuptools==67.4.0 && \
   /opt/venv/bin/python3 -m pip install --no-deps -r requirements/prod.txt -r requirements/test.txt && \
   /opt/venv/bin/python3 -m pip install -e . && \
   /opt/venv/bin/python3 -m pip check
--- a/requirements/prod.txt
+++ b/requirements/prod.txt
@@ -62,7 +62,7 @@ pycparser==2.21
 pygments==2.14.0
 pyinstrument==3.2.0
 pyinstrument-cext==0.2.2
-pymongo==2.9.5  # For json_util in bson
+pymongo==3.0  # For json_util in bson
 Pympler==0.9
 PyNaCl==1.4.0
 python-dateutil==2.8.2


squeaky-pl:
I can't speak for the original Nylas team so not entirely sure to what extent we use bson encoder but yes we at least rely on the way it serializes datetimes, possibly more. Careful examination off all the JSON fields in the db and accompanying code would answer that, and I agree we should not use pymongo.

squeaky-pl (approved):

neob91-close:
I collected all payloads ever returned from `json_util.dumps` in the sync-engine tests as well as in the pymongo bson tests.

For multiple versions of `pymongo` (2.9.5, 3.0 and 3.3), I did the following:
- I checked that loading each payload with `json_util.loads` and dumping it again with `json_util.dumps` results in the initial payload
- I generated an output file that showed the repr(...) of the loaded result

I attach the output files for my tests.

[3.0 output](https://github.com/closeio/sync-engine/files/11229537/test30.txt)
[3.3 output](https://github.com/closeio/sync-engine/files/11229538/test33.txt)
[2.9.5 output](https://github.com/closeio/sync-engine/files/11229539/test295.txt)
[payloads](https://github.com/closeio/sync-engine/files/11229564/dumps.txt)


Diffs:

```diff
--- test295.txt     2023-04-14 08:26:04.000000000 +0200
+++ test33.txt      2023-04-14 08:30:25.000000000 +0200
@@ -969,8 +969,8 @@
 {"id": {"$oid": "6438ee635ec742000b276761"}}
 {'id': ObjectId('6438ee635ec742000b276761')}
 {"r": {"$regex": "a*b", "$options": "iu"}}
-{'r': re.compile('a*b', re.IGNORECASE)}
+{'r': Regex('a*b', re.IGNORECASE|re.UNICODE)}
 {"r": {"$regex": "a*b", "$options": "iu"}}
-{'r': re.compile('a*b', re.IGNORECASE)}
+{'r': Regex('a*b', re.IGNORECASE|re.UNICODE)}
 {"uuid": {"$uuid": "f47ac10b58cc4372a5670e02b2c3d479"}}
 {'uuid': UUID('f47ac10b-58cc-4372-a567-0e02b2c3d479')}
```

```diff
--- test295.txt     2023-04-14 08:26:04.000000000 +0200
+++ test30.txt      2023-04-14 08:29:01.000000000 +0200
@@ -969,8 +969,8 @@
 {"id": {"$oid": "6438ee635ec742000b276761"}}
 {'id': ObjectId('6438ee635ec742000b276761')}
 {"r": {"$regex": "a*b", "$options": "iu"}}
-{'r': re.compile('a*b', re.IGNORECASE)}
+{'r': Regex('a*b', re.IGNORECASE|re.UNICODE)}
 {"r": {"$regex": "a*b", "$options": "iu"}}
-{'r': re.compile('a*b', re.IGNORECASE)}
+{'r': Regex('a*b', re.IGNORECASE|re.UNICODE)}
 {"uuid": {"$uuid": "f47ac10b58cc4372a5670e02b2c3d479"}}
 {'uuid': UUID('f47ac10b-58cc-4372-a567-0e02b2c3d479')}
```

I think it's fairly safe to bump to 3.0 today, and we should be able to bump to 3.3 later.

squeaky-pl:
I think we don't serialize regexes, I didn't see it anywhere in the db or code.

PR #487, opened 2023-04-18
Microsoft: Gracefully handle invalid event organizer

squeaky-pl:
We currently have a loud Rollbar happening in a produciton for a single account,organizer address is supposed to be an email address, but as always there's non-confirming data.

### DIFF
--- a/tests/events/microsoft/test_parse.py
+++ b/tests/events/microsoft/test_parse.py
@@ -1380,6 +1380,16 @@ def test_parse_event_singular():
     assert event.read_only is False
 
 
+def test_parse_event_with_invalid_organizer():
+    event_with_invalid_organizer = single_instance_event.copy()
+    event_with_invalid_organizer["organizer"] = {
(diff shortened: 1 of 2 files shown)


tests/events/microsoft/test_parse.py
```diff
@@ -1380,6 +1380,16 @@ def test_parse_event_singular():
     assert event.read_only is False
 
 
+def test_parse_event_with_invalid_organizer():
+    event_with_invalid_organizer = single_instance_event.copy()
+    event_with_invalid_organizer["organizer"] = {
+        "name": "Garrett’s Bespoke Barber Shop",
```
neob91-close, line 1386: Nit: Could we make this field non-unicode to show that the `address` field is actually causing the error? Or maybe this could be a parametrized test?

neob91-close (approved):

neob91-close (approved):

PR #491, opened 2023-07-03
Expose Google conferenceData

squeaky-pl:
To correctly support Zoom and possible other add-ons that you can install into your Google workspace we need to expose Google's `conferenceData` that contains Zoom related data. Zoom add-on does not add anything to event description or location.

To follow previous sync-engine conventions I use `snake_case` instead of `camelCase`. I also only exposed a subset we need i.e. URIs and conference solution name as we don't immediately need all this data and I don't want to needlessly store it in a table that is quite large already. I normally would not want to expose JSON columns but we already do this in sync-engine and if we ever need to support more add-ons and more data it's gonna be easier (every migration needs to be ran across many databases).

Note that I [explicitly validate the JSON values](https://github.com/closeio/sync-engine/pull/491/files#diff-dee09c7df3a55d14417d06301a49bf48015cf0bde72482498c15d77e3fb45d92R492-R541) because I don't want to end up with arbitrary structure saved in the database by mistake. I chose the limits by looking at historical data. sync-engine uses Text columns for storing JSON data.

Deploy plan:
* [x] Run migration across all the database
* [ ] Deploy

### DIFF
--- a/inbox/events/google.py
+++ b/inbox/events/google.py
@@ -24,7 +24,13 @@ from inbox.events.util import (
 )
 from inbox.models import Account, Calendar
 from inbox.models.backends.oauth import token_manager
-from inbox.models.event import EVENT_STATUSES, ConferenceData, EntryPoint, Event
+from inbox.models.event import (
+    EVENT_STATUSES,
+    ConferenceData,
+    ConferenceSolution,
+    EntryPoint,
+    Event,
+)
 
 CALENDARS_URL = "https://www.googleapis.com/calendar/v3/users/me/calendarList"
 STATUS_MAP = {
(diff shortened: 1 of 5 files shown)


squeaky-pl:
Migration:

```sql
- Running upgrade f9dab5e44c0f -> fe0488decbd1

ALTER TABLE event ADD COLUMN conference_data TEXT(4194304);

UPDATE alembic_version SET version_num='fe0488decbd1' WHERE alembic_version.version_num = 'f9dab5e44c0f';
```

inbox/events/google.py
```diff
@@ -488,6 +490,58 @@ def parse_calendar_response(calendar: Dict[str, Any]) -> Calendar:
...
+MAX_LIST_LENGTH = 10
+STRING_VALIDATORS = [
+    attrs.validators.instance_of(str),
+    attrs.validators.max_len(MAX_STRING_LENGTH),
+]
+
+
+@attrs.frozen(kw_only=True)
+class EntryPoint:
+    uri: str = attrs.field(validator=STRING_VALIDATORS)  # type: ignore
```
squeaky-pl, line 503: mypy barfs if you combine `str` validator with `max_len` but it works in practice.

drewler (approved):
I wonder if some of the extra data (eg: IDs, labels) might be useful in the future...but I agree that we shouldn't sync it yet. Let's wait until they have a use case.

PR #496, opened 2023-07-04
Allow running sync locally without webhooks

squeaky-pl:
Sync-engine has two classes for calendar syncing:
- `EventSync` - poll calendar changes
- `WebhookEventSync` - rely on webhooks to detect calendar changes

both Google and Microsoft providers support both.

When working locally you always want to run event syncing using polling, otherwise you are unable to receive calendar changes (unless you set up some webhook tunnel). Historically I've been always editing code locally to run `EventSync` classes instead of `WebhookEventSync` but I got tired of doing it. I've dried up the event syncing greenlets into `start_evet_sync` method.

Until recently we only had Google event sync so the `FEATURE_FLAGS` switch was called `GOOGLE_PUSH_NOTIFICATIONS` which is also confusing because at least I think of mobile push notifications when I hear push notifications. While respecting the old flag for backward compatibility I also added `WEBHOOKS` there which is clearer IMHO.

When working locally you want to set your environment to something like:

```
  environment:
    FEATURE_FLAGS: ical_autoimport # disable webhooks
    CALENDAR_POLL_FREQUENCY: "30" # poll every 30 secounds instead of default 5 minutes
```

Which I will make a default for our local development setup in another PR.

### DIFF
--- a/inbox/config.py
+++ b/inbox/config.py
@@ -109,11 +109,16 @@ def _update_config_from_env(config, env):
                 raise
 
 
-def _get_local_feature_flags(config):
+def _update_config_from_env_variables(config):
     flags = (
         os.environ.get("FEATURE_FLAGS", "") or config.get("FEATURE_FLAGS", "")
     ).split()
     config["FEATURE_FLAGS"] = flags
+    calendar_poll_frequencey = int(
+        os.environ.get("CALENDAR_POLL_FREQUENCY", "")
+        or config.get("CALENDAR_POLL_FREQUENCY", 300)
+    )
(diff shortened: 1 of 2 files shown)


inbox/config.py
```diff
@@ -109,9 +109,16 @@ def _update_config_from_env(config, env):
                 raise
 
 
-def _get_local_feature_flags(config):
-    flags = config.get("FEATURE_FLAGS", "").split()
+def _update_config_from_env_variables(config):
```
squeaky-pl, line 112: Note that sync-engine historically only supported configuration using config files but this is awkward with Docker and other tools. It will be possible to override those two things with environment variables now.

drewler (approved):

PR #504, opened 2023-08-14
Add MODSEQ explicitely for SmarterMail

squeaky-pl:
We have two accounts in production clusters that produce rather loud Rollbar. They are both hosted on a server called `SmarterMail` and it turns out that it does not implement IMAP according to the spec.

Here's a dump of an IMAP session with this server:

```
* OK IMAP4rev1 SmarterMail
01 CAPABILITY
* CAPABILITY IMAP4rev1 AUTH=CRAM-MD5 AUTH=NTLM AUTH=PLAIN SASL-IR UIDPLUS QUOTA MOVE XLIST CHILDREN ENABLE CONDSTORE
01 OK CAPABILITY completed
02 AUTHENTICATE PLAIN
+
[_REDACTED_]
02 OK PLAIN authentication successful
08 LIST "" *
* LIST (\HasChildren) "." INBOX
[_TRIMMED_ ...]
06 SELECT "INBOX"
* 36 EXISTS
* 0 RECENT
* OK [UNSEEN 2] Message 2 is first unseen
* OK [UIDVALIDITY 1] UIDs valid
* OK [UIDNEXT 37] Predicted next UID
* FLAGS (\Answered \Flagged \Deleted \Seen \Draft \Recent)
* OK [PERMANENTFLAGS (\Answered \Flagged \Deleted \Seen \Draft \Recent \*)] Folder flags
06 OK [READ-WRITE] SELECT completed
08 STATUS "INBOX" (HIGHESTMODSEQ)
* STATUS INBOX (HIGHESTMODSEQ 11600044621828)
08 OK STATUS completed
07 UID FETCH 1:* (FLAGS X-GM-LABELS INTERNALDATE) (CHANGEDSINCE 11600044621828)
* 1 FETCH (FLAGS (\Seen)  INTERNALDATE "25-Jul-2023 13:55:25 -0400" UID 1)
* 2 FETCH (FLAGS (\Seen)  INTERNALDATE "27-Jul-2023 09:50:55 -0400" UID 2)
```

The server claims to support `CONDSTORE` but then when issuing `UID FETCH` at the end of the dump `MODSEQ` is not part of the response.

The spec https://datatracker.ietf.org/doc/html/rfc4551#section-3.3.1 says that it should be added implicitly and contains an example:

```
C: s100 UID FETCH 1:* (FLAGS) (CHANGEDSINCE 12345)
S: * 1 FETCH (UID 4 MODSEQ (65402) FLAGS (\Seen))
S: * 2 FETCH (UID 6 MODSEQ (75403) FLAGS (\Deleted))
S: * 4 FETCH (UID 8 MODSEQ (29738) FLAGS ($NoJunk $AutoJunk
     $MDNSent))
S: s100 OK FETCH completed
```

When I do explicit `UID FETCH 1:* (FLAGS X-GM-LABELS INTERNALDATE MODSEQ) (CHANGEDSINCE 11600044621828)` it does return it.
I also checked a bunch of other servers and they do include `MODSEQ` as mandated by spec.

It would be probably OK to always add `MODSEQ` in the code explicitly but I don't want to risk breaking some other IMAP implementation that don't expect this so I gated it by an if check. Later if I'm sure that nothing is broken I might remove it.

### DIFF
--- a/inbox/crispin.py
+++ b/inbox/crispin.py
@@ -1112,8 +1112,25 @@ class CrispinClient:
     def condstore_changed_flags(
         self, modseq: int
     ) -> Dict[int, Union[GmailFlags, Flags]]:
+        """
+        Fetch flags that changed since the given modseq.
+
+        Note that a well-behaved server should always return the MODSEQ
+        as part of the response as specified in
+        https://datatracker.ietf.org/doc/html/rfc4551#section-3.3.1.
+
+        > When CHANGEDSINCE FETCH modifier is specified,
+        > it implicitly adds MODSEQ FETCH message data item
+
+        However, SmarterMail does not do that, so we add it explicitely.
+        """
+        items = (
+            ["FLAGS", "MODSEQ"]
+            if self.conn.welcome.endswith(b" SmarterMail")
+            else ["FLAGS"]
+        )
         data: Dict[int, Dict[bytes, Any]] = self.conn.fetch(
-            "1:*", ["FLAGS"], modifiers=[f"CHANGEDSINCE {modseq}"]
+            "1:*", items, modifiers=[f"CHANGEDSINCE {modseq}"]
         )
         return {
             uid: Flags(ret[b"FLAGS"], ret[b"MODSEQ"][0] if b"MODSEQ" in ret else None)


inbox/crispin.py
```diff
@@ -1112,8 +1112,25 @@ def idle(self, timeout):
...
+        https://datatracker.ietf.org/doc/html/rfc4551#section-3.3.1.
+
+        > When CHANGEDSINCE FETCH modifier is specified,
+        > it implicitly adds MODSEQ FETCH message data item
+
+        However, SmarterMail does not do that, so we add it explicitely.
+        """
+        items = (
+            ["FLAGS", "MODSEQ"]
+            if self.conn.welcome.endswith(b" SmarterMail")
```
drewler, line 1129: The `if` of shame (for them)

drewler (approved):

PR #505, opened 2023-08-14
Correctly handle None event body for Microsoft calendars

squeaky-pl:
Normally a body is a dictionary that looks like this `{"contentType": "html", "content": "..."}`.
It should not be nullable (at least according to Microsoft) but our experience with Rollbars shows otherwise. Not the first time Microsoft contradicts themselves.

### DIFF
--- a/inbox/events/microsoft/parse.py
+++ b/inbox/events/microsoft/parse.py
@@ -562,6 +562,9 @@ def get_event_description(event: MsGraphEvent) -> Optional[str]:
     Returns:
         Plain text string with all the HTML removed
     """
+    if not event["body"]:
+        return None
+
     content_type = event["body"]["contentType"]
 
     assert content_type in ["text", "html"]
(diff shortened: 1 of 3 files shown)


inbox/events/microsoft/parse.py
```diff
@@ -562,6 +562,9 @@ def get_event_description(event: MsGraphEvent) -> Optional[str]:
     Returns:
         Plain text string with all the HTML removed
     """
+    if not event["body"]:
```
drewler, line 565: This will also return `None` if `event["body"] is an empty string, no? I'm not sure if that could cause any issue, probably not. If it's intentional, maybe we could add a test case.

drewler (approved):

inbox/events/microsoft/parse.py
squeaky-pl: It would be really weird if you could get `None | "" | dict` from Microsoft, but yeah it would be correct to get `None` in such case.

PR #520, opened 2023-10-19
Create dependabot.yml

kris-steinhoff:
Add dependabot config

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+# Docs: https://help.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+
+version: 2
+updates:
+  # Enable version updates for python
+  - package-ecosystem: "pip"
+    directory: "/requirements"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 16
+    rebase-strategy: "disabled"
```
nsaje, line 11: we can get rid of this, this is used to reduce the number of builds but there aren't that many builds on sync-engine

nsaje (approved):

PR #555, opened 2023-11-15
Use send_file to create file download response

squeaky-pl:
[Docs](https://flask.palletsprojects.com/en/3.0.x/api/#flask.send_file).

We should not be encoding those headers manually, instead let Flask do its thing.
Given this code was written more than 10 years ago it does not surprise me the way it was done originally back then.

### DIFF
--- a/tests/imap/data.py
+++ b/tests/imap/data.py
@@ -9,6 +9,8 @@ import flanker
 from flanker import mime
 from hypothesis import strategies as s
 
+MAX_INT_VALUE = (1 << 32) - 1
+
 
 def _build_address_header(addresslist):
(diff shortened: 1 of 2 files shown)


tests/imap/data.py
```diff
@@ -9,6 +9,8 @@
 from flanker import mime
 from hypothesis import strategies as s
 
+MAX_INT_VALUE = (1 << 32) - 1
```
squeaky-pl, line 12: This change id unrelated but removes test flakiness that I saw when opening PR. Sometimes we would get test data that would be outside mysql int range.

kris-steinhoff (approved):

PR #559, opened 2023-11-20
Run static checks on Python 3.8

squeaky-pl:
I've noticed that we were still running static checks on Python 3.7 instead of 3.8 which resulted in some weird mismatches in dependabot PRs.


.circleci/config.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - run:
           name: Prepare Environment
           command: |
-            sudo -E pip install --no-deps -r requirements/lint.txt
+            pip install --no-deps -r requirements/lint.txt
```
squeaky-pl, line 16: Ths new CircleCI Python images work without sudo

.python-version
```diff
@@ -1 +1 @@
-3.8.10
+3.8
```
squeaky-pl, line 1: We run 3.8.10 in the containers but circleci has [only newer 3.8.18 available](https://circleci.com/developer/images/image/cimg/python). It refuses to run python if the version prefix does not match. This is fine as those are static checks only.

drewler (approved):

PR #643, opened 2024-01-29
Do not fetch oversized emails from IMAP

squeaky-pl:
We had this client with an email that had no text or html parts but had 1.4GB MP4 attachment in it. This caused OOM while we try to download the message. We are already rejecting messages over 100MB before parsing them so we could as well just not download them at all.

Deployed this change to do the pod where this customer's email sync is running to confirm this solves the OOM issue that was plaguing us for some time.

image


inbox/crispin.py
```diff
@@ -877,6 +879,17 @@ def uids(self, uids: List[int]) -> List[RawMessage]:
...
                 for _ in range(3):
+                    # We already reject parsing messages bigger than MAX_MESSAGE_BODY_PARSE_LENGTH.
+                    # We might as well not download them at all, save bandwidth, processing time
+                    # and prevent OOMs due to fetching oversized emails.
+                    fetched_size: Dict[int, Dict[bytes, Any]] = self.conn.fetch(
+                        uid, ["RFC822.SIZE"]
+                    )
+                    body_size = int(fetched_size.get(uid, {}).get(b"RFC822.SIZE", 0))
+                    if body_size > MAX_MESSAGE_BODY_PARSE_LENGTH:
+                        log.warning("Skipping fetching of oversized message", uid=uid)
```
tsx, line 890: Would it be possible to fetch message ID (and possibly other useful small-size bits of metadata) for this item and log it alongside the uid? It would seriously help debugging "email not synced to Close" tickets in the future.

inbox/constants.py
```diff
@@ -0,0 +1 @@
+MAX_MESSAGE_BODY_PARSE_LENGTH = 100 * 1024 * 1024  # 100 MB
```
tsx, line 1: No longer "parse" anymore, right? Just a "max message body length"

tsx (approved):
A small nit and a suggestion (separate PR is fine)

inbox/crispin.py
squeaky-pl: Good idea, Will follow up with another PR that fetches headers, parses them and adds some.

PR #655, opened 2024-02-01
Automatic ruff fixes

tsx:
No functional changes, just ran `ruff --fix .`.


squeaky-pl:
Can we get fixes without removing unused app imports as first step? Sync-engine is full of import side effects and my previous attempt at this ended up with broken syncing even the tests passed 😥.

tsx:
This is sad. Sure, I can do one without import changes. Could you please point me to an example of an import side effect so I know what to look for?

nsaje:
@squeaky-pl did we find out why it broke the last time or just reverted?

squeaky-pl:
It was 2,5 years ago so I don't remember well. What I remember was that as soon as I shipped the change removing what it looks like not needed imports all the sync pods in all the clusters would start crash lopping because of some import related problem (something assuming that something else is imported earlier).

Admittedly this might be not the case anymore because the code changed substantially.

But to do this with confidence we would need a test that starts all the pod entry points and see if they started without problem.

In the meantime we could ship a version without removing those imports maybe?

PR #657, opened 2024-02-01
Improve docker build in CI

tsx:
This uses full SHA commit IDs like we did before, and drops the -py3.x part of the tag - only one interpreter is supported/used.

There's also some general cleanup and simplification of the GHA workflow.

### DIFF
--- a/.github/workflows/build-and-release.yaml
+++ b/.github/workflows/build-and-release.yaml
@@ -1,9 +1,7 @@
 name: Build and Release
 
 on:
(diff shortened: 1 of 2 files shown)


squeaky-pl:
@tsx I am seeing some failure related to multi-arch-manifest, can we just drop it since it's not really needed for anything?

squeaky-pl:
This can be closed as it got outdated and superseded by series of other peries.

PR #664, opened 2024-02-08
Google: Pass eventTypes

squeaky-pl:
Google will soon start returning other event types in their APIs that we don't care about. This change ensures that everything keeps working as before.

https://developers.google.com/calendar/api/v3/reference/events/list

```
eventTypes 	string 	Event types to return. Optional. This parameter can be repeated multiple times to return events of different types. The default is ["default", "focusTime", "outOfOffice"].

Acceptable values are:

    "default": Regular events.
    "focusTime": Focus time events.
    "outOfOffice": Out of office events.
    "workingLocation": Working location events.
```

image

### DIFF
--- a/inbox/events/google.py
+++ b/inbox/events/google.py
@@ -132,9 +132,10 @@ class GoogleEventsProvider(AbstractEventsProvider):
         url = "https://www.googleapis.com/calendar/v3/calendars/{}/events".format(
             urllib.parse.quote(calendar_uid)
         )
-        params = {"eventTypes": "default"}
         try:
-            return self._get_resource_list(url, updatedMin=sync_from_time_str, **params)
+            return self._get_resource_list(
+                url, updatedMin=sync_from_time_str, eventTypes="default"
+            )
         except requests.exceptions.HTTPError as exc:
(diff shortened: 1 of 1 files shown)


inbox/events/google.py
```diff
@@ -131,8 +132,9 @@ def _get_raw_events(
         url = "https://www.googleapis.com/calendar/v3/calendars/{}/events".format(
             urllib.parse.quote(calendar_uid)
         )
+        params = {"eventTypes": "default"}
         try:
-            return self._get_resource_list(url, updatedMin=sync_from_time_str)
+            return self._get_resource_list(url, updatedMin=sync_from_time_str, **params)
```
wojcikstefan, line 137: Why the use of kwargs? Is it because the linter would complain about passing `eventTypes="default"` explicitly to the function call?

wojcikstefan (approved):

wojcikstefan:
@squeaky-pl shall we ship it?

inbox/events/google.py
squeaky-pl: That was indeed unnecessary, I changed it.

PR #687, opened 2024-02-27
Event: sync events using generators

squeaky-pl:


squeaky-pl:
Actually need to take a second look at this.

evanpenner1:
> Convert the chain of methods used to sync events into generators. This will help with memory usage as the "upstream" is no longer hoarding them first.
give h8iuim bill

PR #698, opened 2024-03-10
Rename `inbox.basicauth` to `inbox.exceptions`

wojcikstefan:
The comment at the top of this module already said:
> TODO(emfree): this is now legitimately just a grab-bag of nebulous
> exceptions. Rename module and clean up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,9 @@ inbox/test/.cache
 /src
 
 # virtualenv
+venv
+.venv
+virtualenv
 /include
 /local
 bin/py*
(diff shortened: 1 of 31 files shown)


.gitignore
```diff
@@ -26,6 +26,9 @@ inbox/test/.cache
 /src
 
 # virtualenv
+venv
+.venv
+virtualenv
```
wojcikstefan, line 31: FYI I also snuck in a commit ignoring these common virtualenv folder names in git. It seems to me that the convention this repo follows is to have the virtualenv in a `venv` folder, judging from the isort config: https://github.com/closeio/sync-engine/blob/07e939072b53360b9178ca7fec07ccc20103f8d5/pyproject.toml#L1-L2

squeaky-pl (approved):

PR #776, opened 2024-06-03
Use local gevent wheel on aarch64

squeaky-pl:
We are currently building gevent ARM64 wheel on x86_64. Using prebuilt wheel we can speed up CI a lot.

Before:

image

After

image


squeaky-pl:
This is no longer needed with ARM64 runners.

PR #778, opened 2024-06-04
Remove Python 3.9 from the build matrix

squeaky-pl:
We've recently adapted github actions into this project. That work was based on a very old branch from 2.5 years ago we had at the beginning when we were only starting to port to ARM64. This was an experiment. That branch coincided with the time I was porting sync-engine to different versions of Python and it was 3.8 and 3.9 at the time. That's why we have them in matrix now but now Python 3.9 workflow just builds Python 3.8 images so it's defunct, complicates yaml file and only wasting us time and $.

### DIFF
--- a/.github/workflows/build-and-release.yaml
+++ b/.github/workflows/build-and-release.yaml
@@ -49,13 +49,6 @@ jobs:
         include:
           - arch: amd64
             build_name: py3.8
-            dockerhub_push: true
-            python_version: "3.8"
-            run_tests: true
-          - arch: amd64
-            build_name: py3.9
-            dockerhub_push: false
-            python_version: "3.9"
             run_tests: true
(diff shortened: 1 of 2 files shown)


drewler (approved):

evanpenner1:
> We've recently adapted github actions into this project. That work was based on a very old branch from 2.5 years ago we had at the beginning when we were only starting to port to ARM64. This was an experiment. That branch coincided with the time I was porting sync-engine to different versions of Python and it was 3.8 and 3.9 at the time. That's why we have them in matrix now but now Python 3.9 workflow just builds Python 3.8 images so it's defunct, complicates yaml file and only wasting us time and $.

head

evanpenner1:
i need pts

PR #779, opened 2024-06-04
Revert "Upgrade good event and letgreen."

evanpenner1:
Reverts closeio/sync-engine#777


squeaky-pl:
Curious if you seeing any particular problem with these gevent and greenlet versions? I tested this and did not see any problems.

PR #799, opened 2024-06-27
Clean up email parsing

squeaky-pl:
This is a small cleanup that I wanted to do before extracting some of those methods out of the classes. I also added some types.

### DIFF
--- a/inbox/mailsync/backends/imap/common.py
+++ b/inbox/mailsync/backends/imap/common.py
@@ -271,10 +271,10 @@ def create_imap_message(
     )
     new_message = Message.create_from_synced(
         account=account,
-        mid=raw_message.uid,
+        imap_uid=raw_message.uid,
         folder_name=folder.name,
         received_date=raw_message.internaldate,
-        body_string=raw_message.body,
+        body=raw_message.body,
     )
 
     # Check to see if this is a copy of a message that was first created
--- a/inbox/mailsync/backends/imap/common.py
+++ b/inbox/mailsync/backends/imap/common.py
@@ -13,9 +13,10 @@ accounts.
 """
 
 from datetime import datetime
-from typing import Any, List, Set
+from typing import List, Set
 
 from sqlalchemy import bindparam, desc
+from sqlalchemy.orm import Session
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.sql.expression import func
 
@@ -55,7 +56,7 @@ def lastseenuid(account_id, session, folder_id):
 
 
 def update_message_metadata(
-    session, account: Account, message: Message, is_draft: bool
+    session: Session, account: Account, message: Message, is_draft: bool
 ) -> None:
     """Update the message's metadata"""
     # Sort imapuids in a way that the ones that were added later come last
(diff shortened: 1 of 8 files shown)


inbox/mailsync/backends/imap/common.py
```diff
@@ -271,10 +272,10 @@ def create_imap_message(
     )
     new_message = Message.create_from_synced(
         account=account,
-        mid=raw_message.uid,
+        imap_uid=raw_message.uid,
```
squeaky-pl, line 275: Changed this to `imap_uid` because I don't know what `mid` means without reading code.

Is this `middle` like `mid-way` `mid-journey`? No
Is this message id like `Message-ID` header? No
Is this the unique id that IMAP uses to identify a message inside a folder, yes.

inbox/mailsync/backends/imap/common.py
```diff
@@ -271,10 +272,10 @@ def create_imap_message(
...
         folder_name=folder.name,
         received_date=raw_message.internaldate,
-        body_string=raw_message.body,
+        body=raw_message.body,
```
squeaky-pl, line 278: the `body_string` is confusing because `body` is actually always `bytes`

inbox/models/message.py
```diff
@@ -288,67 +288,56 @@ def create_from_synced(
...
-        body_string : bytes
+        body : bytes
             The full message including headers (encoded).
 
         """
-        _rqd = [account, mid, folder_name, body_string]
-        if not all([v is not None for v in _rqd]):
-            raise ValueError(
-                "Required keyword arguments: account, mid, folder_name, body_string"
-            )
```
squeaky-pl, line 303: This is redundant. Never happens in production and not needed with mypy.

inbox/models/message.py
```diff
@@ -288,67 +288,56 @@ def create_from_synced(
...
+        message.namespace_id = account.namespace.id
 
         try:
-            body_length = len(body_string)
+            body_length = len(body)
             if body_length > MAX_MESSAGE_BODY_LENGTH:
                 raise MessageTooBigException(body_length)
-            parsed: MimePart = mime.from_string(body_string)
-            # Non-persisted instance attribute used by EAS.
-            msg.parsed_body = parsed
```
squeaky-pl, line 325: This is only used in `get_headers()` below, that is not used. So I just deleted it.

inbox/models/message.py
```diff
@@ -619,8 +607,6 @@ def _mark_error(self) -> None:
         self.size = 0
         if self.received_date is None:
             self.received_date = datetime.datetime.utcnow()
-        if self.body is None:
-            self.body = None
```
squeaky-pl, line 623: This is just 🤡

inbox/models/message.py
```diff
@@ -757,12 +743,6 @@ def attached_event_files(self) -> List["Part"]:
     def account(self) -> Account:
         return self.namespace.account
 
-    def get_header(self, header: str, mid: int) -> Optional[str]:
-        if self.decode_error:
-            log.warning("Error getting message header", mid=mid)
-            return None
-        return self.parsed_body.headers.get(header)
-
```
squeaky-pl, line 765: dead code

drewler (approved):

PR #821, opened 2024-07-29
Run arm64 on gh-arm64

squeaky-pl:
Thanks to devops team we can now run sync-project Github Actions on ARM64, since ARM64 hosted runners are now [available](https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/) and our devops team enabled them. Note that we need to opt-in open source repositories one by one and that needs devops intervention. We don't allow running arbitrary open source repositories on ARM64 hosted runners.

Previously we were running ARM builds on x86-64 with qemu emulation, this was very slow.

After this change ARM builds go from ~30 minutes to ~5 minutes. We are also able to reuse the same workflow instead of having a separate one for ARM so this simplifies things as well.


.github/workflows/build-and-release.yaml
```diff
@@ -110,6 +109,7 @@ jobs:
           else
             echo Skipping tests
           fi
+        if: ${{ matrix.arch == 'amd64' }}
```
squeaky-pl, line 112: We run tests only on x86

drewler (approved):

PR #827, opened 2024-07-31
Sync pods: Use greenlet.kill() instead of gevent.kill(greenlet)

squeaky-pl:
`greenlet.kill()` and `gevent.kill(greenlet)` do exactly the same thing.

By using OOO version we get get rid of `gevent` APIs. Python threads don't support `thread.kill()` but later when we switch to threads I'll implement a polyfill for this that emulates greenlet behavior as close as possible.

### DIFF
--- a/inbox/mailsync/service.py
+++ b/inbox/mailsync/service.py
@@ -359,14 +359,14 @@ class SyncService:
                 return False
         return True
 
-    def stop(self, *args):
+    def stop(self) -> None:
         self.log.info("stopping mail sync process")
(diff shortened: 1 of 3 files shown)


inbox/mailsync/service.py
```diff
@@ -359,14 +359,14 @@ def start_sync(self, account_id: int) -> bool:
                 return False
         return True
 
-    def stop(self, *args):
+    def stop(self) -> None:
```
squeaky-pl, line 362: args was unused

drewler (approved):

PR #829, opened 2024-07-31
Revert "Sync pods: Use greenlet.kill() instead of gevent.kill(greenlet) (#827)"

squeaky-pl:
I discovered in production that after merging reverted PR the shutdown time of sync containers has become much slower (20s versus almost 5 minutes).

So there's for sure a difference in practice between the two methods of killing the greenlets or some other factor that I do not know about yet that plays important role here.

I will investigate separately, for now let's revert it.


squeaky-pl:
Superseded by https://github.com/closeio/sync-engine/pull/830

drewler (approved):

PR #830, opened 2024-07-31
Bugfix: Fix signal handler

squeaky-pl:
In [here](https://github.com/closeio/sync-engine/pull/827/files) I removed what I though was unused argument. But it was used to consume signal handler parameters...

As a result of that change the signal handler is not working anymore and the process won't stop if you send it SIGTERM or SIGINT. Instead it will be killed by k8s after ~5 minutes. I realized that because deploys have become much after this change.

That's what I got for sneaking unrelated changes 🤡

I chose to consume the arguments in another function, because consuming them inside a method on a greenlet is confusing IMHO.

### DIFF
--- a/bin/inbox-start.py
+++ b/bin/inbox-start.py
@@ -10,6 +10,7 @@ import platform
 import signal
 import socket
 import sys
+import types
 
 import click
 import setproctitle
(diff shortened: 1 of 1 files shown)


squeaky-pl:
This actually works locally but doesn't fix all of it in production 🤔 let's revert the original PR for now.

squeaky-pl:
Superseded by https://github.com/closeio/sync-engine/pull/921

PR #831, opened 2024-08-01
Sync pods: Replace gevent.queue.Queue with queue.Queue

squeaky-pl:
Sync-engine calls gevent.monkey.patch_all() before importing other modules.

`gevent.queue.Queue` APIs are a superset of stdlib `queue.Queue` APIs.

Sync-engine uses `items` convenience parameter of `gevent.queue.Queue` [initializer](https://github.com/gevent/gevent/blob/master/src/gevent/queue.py#L129) to prepulate queue elements. `queue.Queue` [initializer](https://github.com/python/cpython/blob/main/Lib/queue.py#L46) does not have this argument, but you can just put them right after initialization.

The tests rely on the fact that `gevent.queue.Queue` implements [iterator protocol](https://github.com/gevent/gevent/blob/a67891991c95b5ffca78dc1f1a11f956ec3963ca/src/gevent/queue.py#L408-L415) and stdlib does not, but you can just pop the elements to a list instead and check them.

### DIFF
--- a/inbox/crispin.py
+++ b/inbox/crispin.py
@@ -47,13 +47,13 @@ imaplib.InternalDate = re.compile(  # type: ignore
 )
 
 import functools
+import queue
 import threading
 from collections import defaultdict, namedtuple
 from email.parser import HeaderParser
 from threading import BoundedSemaphore
 
 from gevent import socket
-from gevent.queue import Queue
 from sqlalchemy.orm import joinedload
 
 from inbox.exceptions import GmailSettingError
@@ -228,7 +228,9 @@ class CrispinConnectionPool:
         )
         self.account_id = account_id
         self.readonly = readonly
-        self._queue = Queue(num_connections, items=num_connections * [None])
+        self._queue: "queue.Queue[CrispinClient | None]" = queue.Queue(num_connections)
+        for _ in range(num_connections):
+            self._queue.put(None)
         self._sem = BoundedSemaphore(num_connections)
         self._set_account_info()
 
--- a/tests/imap/test_pooling.py
+++ b/tests/imap/test_pooling.py
@@ -1,6 +1,8 @@
 import imaplib
+import queue
 import socket
 import ssl
+from typing import Any, List
 from unittest import mock
 
 import gevent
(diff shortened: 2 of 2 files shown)


drewler (commented):
Have you checked if this actually makes `gevent.queue.Queue` available when `import queue`? I remember some whackiness about `queue` patching and `gevent`, and [`gevent.monkey.patch_queue` seems to just patch `SimpleQueue`](https://github.com/gevent/gevent/blob/229a1b3747e88f823700538a74a685b75303fc53/src/gevent/monkey.py#L542-L553).

```python
from gevent.monkey import patch_all

patch_all()

import threading
import queue

threading.Thread # <class 'gevent.threading.Thread'>
queue.Queue # <class 'queue.Queue'>


```

squeaky-pl:
My bad for not explaining this correctly. The `queue.Queue` *does not need* to be patched to cooperate with gevent, because it's internals are patched instead.

If you look here you can see that there's a `threading.Lock` [here](https://github.com/python/cpython/blob/main/Lib/queue.py#L54) wrapped in a `threading.Condition`s [here](https://github.com/python/cpython/blob/main/Lib/queue.py#L56-L66)

Then `.put()` and `.get()` uses those [here](https://github.com/python/cpython/blob/88030861e216ac791725c8784752201d6fe31329/Lib/queue.py#L153-L178) and [here](https://github.com/python/cpython/blob/main/Lib/queue.py#L194-L218).

`threading.Condition` is a wrapper around RLock which is an alias for _RLock which is powered by `_thread.allocate_lock()`as seen [here](https://github.com/python/cpython/blob/88030861e216ac791725c8784752201d6fe31329/Lib/threading.py#L269-L307)

The underlying primitives get patched

```
>>> from gevent.monkey import patch_all
>>> 
>>> patch_all()
True
>>> import threading
>>> threading.Lock
<class 'gevent.thread.LockType'>
>>> import _thread
>>> _thread.allocate_lock
<class 'gevent.thread.LockType'>
```

You can also confirm it empirically with a following program


```python
from gevent.monkey import patch_all
patch_all()

def block():
    import queue
    queue = queue.Queue()
    # this blocks forever since the queue is empty
    queue.get()


def printer():
    import time
    while True:
        print("alive")
        time.sleep(1)


import gevent
printer_greenlet = gevent.spawn(printer)
block_greenlet = gevent.spawn(block)
gevent.joinall([printer_greenlet, block_greenlet])
```


It would stop printing pretty quickly if `queue.Queue` didn't cooperate with gevent but it does work

```
sync-engine@8d923f4bb489:/opt/app$ python3 x.py
alive
alive
alive
alive
alive
alive
alive
alive
alive
alive
alive
alive
```

drewler (approved):
I see. Thanks for the explanation!

PR #847, opened 2024-08-06
query optimization

squeaky-pl:
- Pass lastseenuid into self.get_new_uids if avaialble
- In progress


squeaky-pl:
Closing this, as not worth pursuing as it substantially worsened memory usage and complicated code

PR #771, opened 2024-05-29
Remove pymongo dependency from this project

squeaky-pl:
The original authors of this project chose to use pymongo's bson utils to handle JSON serialization and deserialization into various VARCHAR types in MySQL.

We have an unresolved security issue related to pymongo.

image

I've had questions in the past why are we using pymongo here, it suggests that sync-engine uses MongoDB while it does not. I think it's time to break that dependency because it just brings confusion, and pymongo will continue to receive security patches often as it contains C code which makes it prone to security bugs.

MongoDB's extended JSON syntax can serialize more types than stdlib's json module. [The docs](https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/).

Pymongo's implementation sits in here https://github.com/mongodb/mongo-python-driver/blob/18328a909545ece6e1cd7e172e28271a59e367d5/bson/json_util.py. As you can see it handles many extra types. I analyzed all the call sites and data in those columns for `\$\w` references and concluded that we are only using `"field": {"$date": <timestmp>}` syntax in sync-engine.

Example

```json
{"sync_disabled_by": null, "sync_start_time": {"$date": 1717004873429}, "original_start_time": {"$date": 1522779878389}, "sync_error": null, "sync_disabled_on": null, "sync_end_time": null, "sync_disabled_reason": null}
```

I've vendored the code and then stripped it to only include the parts that we are using.

Separately I think it makes sense to [completely get rid of this $date syntax](https://github.com/closeio/sync-engine/issues/775) as I don't find it particularly readable.

I've already tested those changes first on my account in production and also on an entire cluster (nylas-e) and did not see any problems.

<details>
<summary>My intenral monologue about columns using JSON types</summary>

* `actionlog.extra_args` https://github.com/closeio/sync-engine/blob/64e70ec06139ab63f676c98c51db9eaccfbbbe6d/inbox/models/action_log.py#L67

```sql
SELECT DISTINCT extra_args FROM (SELECT extra_args from actionlog a LIMIT 500000) As a;
```

|extra_args|
|----------|
|{"unread": false}|
|{"added_labels": [], "removed_labels": ["\\Inbox"]}|
|{"destination": "INBOX.Archive"}|
|{"destination": "INBOX"}|
|{"added_labels": ["\\Trash"], "removed_labels": ["\\Inbox"]}|
|{"added_labels": ["\\Inbox"], "removed_labels": []}|
|{"added_labels": [], "removed_labels": ["\\Inbox", "\\Trash"]}|
[...]

PR #863, opened 2024-08-22
Remove packages that were needed to build pymongo

squeaky-pl:
Sync-engine used to depend on pymongo, but I [removed that dependency](https://github.com/closeio/sync-engine/pull/771).

Previously I needed distutils and pinning setuptools version so I could build pymongo wheel, but since pymongo is no longer here we don't need those.


drewler (approved):
✨🧹

wojcikstefan:
Yay!!! ⭐

PR #864, opened 2024-08-22
Slim down docker images

squeaky-pl:
I was revising the Docker image to realize that we have quite a number of packages that are not needed. Most of them used to be build time dependencies of Python packages that can be now installed from binary wheels.

This:

- Removes dependencies we don't use, so there's less of security updates needed
- Makes the images smaller ( 1.12GB -> 896MB )
- Makes the builds faster

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,37 +3,33 @@ FROM ubuntu:20.04
 RUN groupadd -g 5000 sync-engine \
   && useradd -d /home/sync-engine -m -u 5000 -g 5000 sync-engine
 
-ENV TZ="Etc/GMT"
-ENV DEBIAN_FRONTEND=noninteractive
+ENV \
+  TZ="Etc/GMT" \
+  LANG="en_US.UTF-8" \
+  LC_ALL="en_US.UTF-8" \
+  DEBIAN_FRONTEND=noninteractive \
+  PATH="/opt/venv/bin:$PATH"
+
 ARG BUILD_WEEK=0
 RUN echo $BUILD_WEEK && apt-get update \
   && apt-get dist-upgrade -y \
-  && apt-get install -y \
+  && apt-get install --no-install-recommends -y \
     tzdata \
-    build-essential \
+    locales \
+    make \
     curl \
-    dnsutils \
+    gpg \
+    gpg-agent \
+    dirmngr \
     gcc \
-    g++ \
     git \
     python3-dev \
     python3-pip \
-    wget \
     gettext-base \
-    language-pack-en \
-    libcurl4-openssl-dev \
     libmysqlclient-dev \
-    libxml2-dev \
-    libxslt-dev \
-    libxslt1-dev \
     mysql-client \
-    pkg-config \
-    lsof \
-    net-tools \
-    shared-mime-info \
-    telnet \
     vim \
-    libffi-dev \
+  && locale-gen en_US.UTF-8 \
   && rm -rf /var/lib/apt/lists/*
 
 RUN mkdir /etc/inboxapp && \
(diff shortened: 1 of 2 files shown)


requirements/prod.txt
```diff
@@ -11,7 +11,6 @@ jedi==0.18.2
 boto==2.49.0
 boto3==1.19.4
 botocore==1.22.9
-cchardet==2.1.7
```
squeaky-pl, line 14: cchardet is a faster version of chardet. We don't often need to detect charset on the fly and if we need to we can just use chardet. cchardet Requires a C++ compiler and it used to be an optional dependency of requests. It is abandoned - see the [drama here](https://github.com/PyYoshi/cChardet/issues/77) and everybody moved away from it.

Dockerfile
```diff
@@ -3,37 +3,32 @@ FROM ubuntu:20.04
...
   && useradd -d /home/sync-engine -m -u 5000 -g 5000 sync-engine
 
-ENV TZ="Etc/GMT"
-ENV DEBIAN_FRONTEND=noninteractive
+ENV \
+  TZ="Etc/GMT" \
+  LANG="en_US.UTF-8" \
+  LC_ALL="en_US.UTF-8" \
+  DEBIAN_FRONTEND=noninteractive \
+  PATH="/opt/venv/bin:$PATH"
```
squeaky-pl, line 11: Just made one layer from all the env vars.

Dockerfile
```diff
@@ -3,37 +3,33 @@ FROM ubuntu:20.04
...
-    libxslt1-dev \
     mysql-client \
-    pkg-config \
-    lsof \
-    net-tools \
-    shared-mime-info \
-    telnet \
     vim \
-    libffi-dev \
+  && locale-gen en_US.UTF-8 \
```
squeaky-pl, line 32: We used to install `language-pack-en`, that just includes all the English variants around the world, but we only need `en_US.UTF-8`

Dockerfile
```diff
@@ -3,37 +3,33 @@ FROM ubuntu:20.04
...
+    gpg-agent \
+    dirmngr \
     gcc \
-    g++ \
     git \
     python3-dev \
     python3-pip \
-    wget \
     gettext-base \
-    language-pack-en \
```
squeaky-pl, line 23: that was substituted with `locale-gen en_US.UTF-8`

Dockerfile
```diff
@@ -49,16 +45,10 @@ USER sync-engine
 
 WORKDIR /opt/app
 
-ENV PATH="/opt/venv/bin:$PATH"
-
 COPY --chown=sync-engine:sync-engine ./ ./
 RUN python3 -m pip install pip==24.0 virtualenv==20.25.1 && \
   python3 -m virtualenv /opt/venv && \
-  /opt/venv/bin/python3 -m pip install --no-deps -r requirements/prod.txt -r requirements/test.txt && \
+  /opt/venv/bin/python3 -m pip install --no-cache --no-deps -r requirements/prod.txt -r requirements/test.txt && \
```
squeaky-pl, line 51: We don't need to store wheels in the image since they will be only used once for install and it increases image size.

Dockerfile
```diff
@@ -3,37 +3,33 @@ FROM ubuntu:20.04
...
+  && apt-get install --no-install-recommends -y \
     tzdata \
-    build-essential \
+    locales \
+    make \
     curl \
-    dnsutils \
+    gpg \
+    gpg-agent \
+    dirmngr \
```
squeaky-pl, line 23: Those were previously installed from recommended, we need those for some things devops uses in the upstream image.

drewler (approved):

PR #868, opened 2024-08-26
Multistage Docker

squeaky-pl:
This was prompted by an security alert from our security solution. It's claiming that linux package needs an update:

Screenshot 2024-08-26 at 15 45 07

but querying dpkg shows that no `linux` package is present:

```
sync-engine@c1d3b7d8cc12:/opt/app$ dpkg -l | grep linux
ii  binutils-aarch64-linux-gnu 2.34-6ubuntu1.9                   arm64        GNU binary utilities, for aarch64-linux-gnu target
ii  libselinux1:arm64          3.0-1build2                       arm64        SELinux runtime shared libraries
ii  linux-libc-dev:arm64       5.4.0-193.213                     arm64        Linux Kernel Headers for development
ii  util-linux                 2.34-0.1ubuntu9.6                 arm64        miscellaneous system utilities
```

Nonetheless you can see that `linux-libc-dev` is present with the version matching the alert.

This is because unlike in other projects we are still using single stage Dockerfile. We can get rid of this packages by splitting the dockerfile to 3 stages like elsewhere. This way any Python package build time dependencies are not present in the final image.

```
sync-engine@9266855b4611:/opt/app$ dpkg -l | grep linux
ii  libselinux1:arm64          3.0-1build2                       arm64        SELinux runtime shared libraries
ii  util-linux                 2.34-0.1ubuntu9.6                 arm64        miscellaneous system utilities
```

The additional perk of multistage Dockerfile is reduced image size: from 896MB to 581MB.
Combined with the savings from [this PR](https://github.com/closeio/sync-engine/pull/864) we managed to make the image 2 times smaller.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:20.04 AS stage_0
 
 RUN groupadd -g 5000 sync-engine \
   && useradd -d /home/sync-engine -m -u 5000 -g 5000 sync-engine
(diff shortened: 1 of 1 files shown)


mrhiggi-close (approved):
Cutting the image size down 2x is awesome 🥇

PR #874, opened 2024-09-04
Script to remove existing message attachments

squeaky-pl:
As noted in the PR in the downstream we were storing decoded attachments all this time, but we were not using them since CRM parses the raw MIME. After setting `STORE_MESSAGE_ATTACHMENTS` to False, it's time to delete all this cruft. 

Note that sync-engine does not follow the architecture we use anywhere else.

This has a service-repo-like function called `find_blocks` that basically finds all the attachments sha256. This is done in batches so it doesn't hold to MySQL transactions forever.

And then there's main function that either deletes things or says it would delete them based on `--dry-run/--no-dry-run`

This outputs something like this, mainly for the sake of me seeing that it does something

```
1/10 2024-07-30 would-delete 1eb8021f201310b1d679dae7bf1bd1e658a11ada2a4a9093f23f72a79a911661 1407 1
3/10 2024-07-30 would-delete 598642c91e088f3ff89ee1105b6591dad1ba73216f4a73069bfa0f9894bc1907 2054 1
5/10 2024-07-30 would-delete b7fcf6086d83374e4ecd99f8b4d05c7ea734ec78db2bfabda86ffed234b37fa8 1162 1
7/10 2024-07-30 would-delete e35031e8cb85615f25d36e572fd5c92b3581aa44407387356015a98fd65e84ea 1809 1
...
10/10 2024-07-30 would-delete f5cdf4fd04d2ab76c716d78a872ba69172e1b3d4eadc52ca9c64e2cb0ebc2481 57366 1
```

where the columns are `id/maximum id date-created resolution sha256 size number-of-parts` but this is not really important.


bin/remove-message-attachments.py
```diff
@@ -0,0 +1,127 @@
...
+) -> None:
+    blocks = find_blocks(
+        limit,
+        datetime.datetime.fromisoformat(after) if after else None,
+        datetime.datetime.fromisoformat(before) if before else None,
+    )
+
+    for block, max_id in blocks:
+        if check_existence:
+            data = blockstore.get_from_blockstore(block.data_sha256)
```
squeaky-pl, line 104: We don't really need to check if the key exists before deleting it. The `blockstore.delete_from_blockstore` will happily accept non existent keys. Downloading the data slows the process down of course, but I have `--check-existance` that is `False` by default to see if the keys are present before deleting it just in case I want it.

bin/remove-message-attachments.py
```diff
@@ -0,0 +1,126 @@
...
+        else:
+            data = ...  # assume it exists, it's OK to delete non-existent data
```
mrhiggi-close, line 105: Neat ellipsis

mrhiggi-close (approved):
👍

PR #878, opened 2024-09-05
Compress RAW MIMEs with zstandard

squeaky-pl:
As presented during backend-sync zstandard can compress emails very well.

- For binary attachments like png/jpegs... it can reduce their size by 30% due to the fact that they are base64 encoded in MIME
- For longer English texts the compression can get as good as 5 times due the repetition of HTML and text parts that contain the same words

This lands the code and tests that use zstandard before saving the raw MIME data to blockstores.

In this PR I am not turning the compression on yet. This is controlled by config variable so this changes nothing so far.
There will be a separate PR to turn on `COMPRESS_RAW_MIME` in the downstream repository.
There will be also a separate PR with a script to recompress all the existing raw mimes.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,6 @@
 * text eol=lf
+*.eml eol=crlf
 *.png binary
 *.jpg binary
 *.ics binary
+
--- a/inbox/util/blockstore.py
+++ b/inbox/util/blockstore.py
@@ -19,6 +19,10 @@ from boto.s3.connection import S3Connection
 from boto.s3.key import Key
 
 # https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#zstandard-frames
+# > This value was selected to be less probable to find at the beginning of some random file.
+# > It avoids trivial patterns (0x00, 0xFF, repeated bytes, increasing bytes, etc.),
+# > contains byte values outside of ASCII range, and doesn't map into UTF8 space.
+# > It reduces the chances that a text file represent this value by accident.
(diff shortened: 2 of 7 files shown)


.gitattributes
```diff
@@ -1,4 +1,6 @@
 * text eol=lf
+*.eml eol=crlf
```
squeaky-pl, line 2: This is needed because otherwise git would change `\r\n\` to `\n` as stated in the line above. Emails (as HTTP does) always use Windows `\r\n` CRLF always but they are text files.

inbox/util/blockstore.py
```diff
@@ -97,10 +150,44 @@ def get_from_blockstore(data_sha256) -> Optional[bytes]:
         log.warning("No data returned!")
         return value
 
+    if check_sha:
+        assert (
+            data_sha256 == sha256(value).hexdigest()
+        ), "Returned data doesn't match stored hash!"
```
squeaky-pl, line 156: This is added here because previously we would check hash here, there was no compression involved before. This is still `True` by default but when retrieving MIME we have to decompress it before checking the hash and this is now done in `get_raw_mime` below.

tests/util/test_blockstore.py
```diff
@@ -0,0 +1,44 @@
+import hashlib
+import pathlib
+
+import pytest
+
+from inbox.util import blockstore
+
+
+@pytest.mark.usefixtures("blockstore_backend")
+@pytest.mark.parametrize("blockstore_backend", ["disk", "s3"], indirect=True)
```
squeaky-pl, line 10: sync-engine has two blockstore backends, S3 and disk. We use S3 in production but disk in our local devtools setup.

PR #901, opened 2024-09-24
Gunicorn JSON logs, the universal way

squeaky-pl:
Adapted from https://til.codeinthehole.com/posts/how-to-get-gunicorn-to-log-as-json/.

Previously the way logging was done was hacky, i.e. we were subclassing internals of the gevent worker to make it work. This is not portable between different worker implementations (i.e. thread and gevent) and the implementation is just ugly. Gunicorn has a standard logging syntax described here https://docs.gunicorn.org/en/stable/settings.html#access-log-format that works between all of them. Also note that Gunicorn is not really designed to use structlog, it's just easier to bite the bullet and format stdlib logging with one of the JSON formatters like `json_log_formatter`.

I already deployed this to single workers to confirm that everything keeps working as expected.

The actual logdict configuration needs to go to the downstream as we maintain Gunicorn configuration there:

```
# Ensure the two named loggers that Gunicorn uses are configured to use a custom
# JSON formatter.
logconfig_dict = {
    "version": 1,
    "formatters": {
        "json_request": {
            "()": JsonRequestFormatter,
        },
        "json_error": {
            "()": JsonErrorFormatter,
        },
    },
    "handlers": {
        "json_request": {
            "class": "logging.StreamHandler",
            "stream": sys.stdout,
            "formatter": "json_request",
        },
        "json_error": {
            "class": "logging.StreamHandler",
            "stream": sys.stdout,
            "formatter": "json_error",
        },
    },
    "root": {"level": "INFO", "handlers": []},
    "loggers": {
        "gunicorn.access": {
            "level": "INFO",
            "handlers": ["json_request"],
            "propagate": False,
        },
        "gunicorn.error": {
            "level": "INFO",
            "handlers": ["json_error"],
            "propagate": False,
        },
    },
}
```

I will open another PR there.

### DIFF
--- a/inbox/api/wsgi.py
+++ b/inbox/api/wsgi.py
@@ -2,12 +2,10 @@ import gevent.monkey
 
 gevent.monkey.patch_all()
 
-import errno
-import socket
+import logging
 import sys
 
-import gunicorn.glogging
-from gevent.pywsgi import WSGIHandler, WSGIServer
+import json_log_formatter
 from gunicorn.workers.ggevent import GeventWorker
 
 from inbox.instrumentation import Tracer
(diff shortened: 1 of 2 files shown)


inbox/api/wsgi.py
```diff
@@ -16,130 +14,81 @@
...
     gunicorn.workers.ggevent.GeventPyWSGIWorker.
     """
 
-    server_class = WSGIServer
-    wsgi_handler = NylasWSGIHandler
-
     def init_process(self):
         print("Python", sys.version, file=sys.stderr)
 
+        maybe_enable_rollbar()
```
squeaky-pl, line 25: I rolled `RollbarWSGIWorker` into this class because it was useless inheritance.

inbox/api/wsgi.py
```diff
@@ -16,130 +14,81 @@
...
+
+        return dict(
+            response_bytes=record.args["B"],
+            request_time=float(record.args["L"]),
+            remote_address=record.args["h"],
+            http_status=record.args["s"],
+            http_request=f"{method} {url}",
+            request_method=method,
+            **log_context,
+        )
```
squeaky-pl, line 72: This maintains exactly the same access log format as we had in the previous implementation.

inbox/api/wsgi.py
```diff
@@ -16,130 +14,81 @@
...
+    ) -> "dict[str, str | int | float]":
+        # Convert the log record to a JSON object.
+        # See https://docs.gunicorn.org/en/stable/settings.html#access-log-format
+
+        url = record.args["U"]
+        if record.args["q"]:
+            url += f"?{record.args['q']}"
+
+        method = record.args["m"]
+        log_context = record.args.get("{log_context}e", {})
```
squeaky-pl, line 62: Yes `{log_context}e`, where `e` stands for WSGI environment. These are "extras" we attach on the fly in the Flask views.

inbox/api/wsgi.py
```diff
@@ -16,130 +14,81 @@
...
 
-class NylasWSGIHandler(WSGIHandler):
-    """Custom WSGI handler class to customize request logging. Based on
-    gunicorn.workers.ggevent.PyWSGIHandler.
-    """
-
-    def log_request(self):
-        # gevent.pywsgi tries to call log.write(), but Python logger objects
-        # implement log.debug(), log.info(), etc., so we need to monkey-patch
-        # log_request(). See
```
squeaky-pl, line 27: 🤮

drewler (approved):

PR #902, opened 2024-09-24
Drop gevent in API pods #2

squeaky-pl:
After PR #902 it becomes fairly easier to switch API pods from gevent to threads:

- Remove monkeypatching
- Change the base class of the worker to the threading one
- Change logging to log thread id instead of greenlet id
- Remove Tracer that detects stuck greenlets (which is greenlet specific). Threads switch by themself every N instructions and whenever C extensions release GIL - unlike greenlets which need to do collaborative I/O call or use some gevent APIs that yield to the event loop

# Based on PR #901

## Full chain of PRs as of 2024-09-24

* PR #902:
  `drop-gevent-api-2` ➔ `gunicorn-json-logs`
* PR #901:
  `gunicorn-json-logs` ➔ `master`


inbox/api/wsgi.py
```diff
@@ -26,7 +22,7 @@ def init_process(self):
 
         configure_logging(log_level=LOGLEVEL)
 
-        if MAX_BLOCKING_TIME:
+        if config.get("USE_GEVENT", True) and MAX_BLOCKING_TIME:
```
mrhiggi-close, line 25: What does it mean if `USE_GEVENT` is True, but we're in a `ThreadWorker` and gevent monkey patches are no longer applied?

squeaky-pl: @mrhiggi-close I removed Tracer entirely (since it's only used from API pods) and also ensured that the invalid combination is now impossible in [`c617f97` (#902)](https://github.com/closeio/sync-engine/pull/902/commits/c617f97ec12250f4b90eacf4cad4b2c7a0d4a6bc)

inbox/logging.py
```diff
@@ -209,7 +210,12 @@ class BoundLogger(structlog.stdlib.BoundLogger):
     """BoundLogger which always adds greenlet_id and env to positional args"""
 
     def _proxy_to_logger(self, method_name, event, *event_args, **event_kw):
-        event_kw["greenlet_id"] = id(gevent.getcurrent())
+        from inbox.config import config
+
+        if config.get("USE_GEVENT", True):
+            event_kw["greenlet_id"] = id(gevent.getcurrent())
+        else:
+            event_kw["thread_id"] = threading.get_ident()
```
squeaky-pl, line 218: Note that this piece is shared between API and sync pods, sync pods still use gevent at the time of writing.

mrhiggi-close (approved):

PR #911, opened 2024-09-30
Synback: Remove gevent

squeaky-pl:
Stacked on: https://github.com/closeio/sync-engine/pull/944

Remove gevent from syncback using interruptible threads. As you can see the diff is minimal. I already tested this end to end.

This also removes gevent from the test suite because we are at the point it can be done.


mrhiggi-close (approved):
Yay gevent dies 🎉

PR #912, opened 2024-09-30
Tests: Remove gevent APIs

squeaky-pl:
Removes gevent-specific APIs from tests. Tests still use gevent through monkey-patching in conftest.py though. I can only remove that later when all the pods are ported away from gevent.


tests/security/test_smtp_ssl.py
squeaky-pl, line 1: This file only contains fixtures that are not used anywhere.

inbox/crispin.py
```diff
@@ -258,7 +262,11 @@ def get(self):
         # The queue implementation does not have that property; having
         # greenlets simply block on self._queue.get(block=True) could cause
         # individual greenlets to block for arbitrarily long.
-        self._sem.acquire()
+
+        succeeded = self._sem.acquire(timeout=timeout)
+        if not succeeded:
+            raise ConnectionPoolTimeoutError()
+
```
squeaky-pl, line 269: Those chages are explained later in the test.

tests/general/test_namespace.py
```diff
@@ -189,8 +188,7 @@ def test_deletion_no_throttle(db, patch_requests_no_throttle):
     db.session.commit()
 
     to_delete = get_accounts_to_delete(0)
-    greenlet = gevent.spawn(batch_delete_namespaces, to_delete, throttle=True)
-    greenlet.join()
```
squeaky-pl, line 193: Those just didn't make sense at all. This spawns a greenlet with a function and then immediately waits for it to end. Logically it's exactly the same as running the function in the same greenlet/thread.

tests/imap/test_actions.py
```diff
@@ -21,10 +20,8 @@
     update_label,
 )
 from inbox.crispin import writable_connection_pool
-from inbox.events.actions.backends.gmail import remote_create_event
 from inbox.models import ActionLog, Category
 from inbox.models.action_log import schedule_action
-from inbox.models.session import new_session
```
squeaky-pl, line 27: Some dead imports, I removed at the same time.

tests/imap/test_pooling.py
```diff
@@ -34,11 +33,10 @@ def test_pool():
...
 
-@pytest.mark.skipif(True, reason="randomly blocks forever")
-def test_block_on_depleted_pool():
+def test_timeout_on_depleted_pool():
     pool = TestableConnectionPool(1, num_connections=1, readonly=True)
-    # Test that getting a connection when the pool is empty blocks
-    with pytest.raises(gevent.hub.LoopExit), pool.get(), pool.get():
+    # Test that getting a connection when the pool is empty times out
+    with pytest.raises(ConnectionPoolTimeoutError), pool.get(), pool.get(timeout=0.1):
         pass
```
squeaky-pl, line 40: Previously this relied on an implementation detail of gevent. `gevent.hub.LoopExit` is internal gevent exception that you are not supposed anywhere because it's used for internal gevent signaling.

The test here wants to test that:
- We have a pool with 1 connection
- We try to get 2 connections, which would block

To test this without relaying on gevent details I added `timeout` parameter. Without `timeout` the call blocks forever. With a timeout when the pool is deplated it will raise an exception `ConnectionPoolTimeoutError` which is not gevent specific.

tests/security/test_smtp_ssl.py
squeaky-pl, line 1: This file contained only unused fixtures.

PR #915, opened 2024-10-01
Remove gevent from inbox.util.concurrency

squeaky-pl:


inbox/util/concurrency.py
```diff
@@ -72,11 +71,10 @@ def wrapped(*args, **kwargs):
         while True:
             try:
                 return func(*args, **kwargs)
-            except gevent.GreenletExit:
-                # GreenletExit isn't actually a subclass of Exception.
-                # This is also considered to be a successful execution
-                # (somebody intentionally killed the greenlet).
-                raise
```
squeaky-pl, line 79: This is just stupid as it catches the exception and then immediately raises it. I left the comment for clarity but the code does nothing useful.

drewler (approved):

PR #916, opened 2024-10-01
Scripts: Remove gevent

squeaky-pl:
This ports all the scripts that we are not running in k8s pods.

While porting and testing I found out that some of them were already broken before and I fixed them.

Note that we have to pass `config["USE_GEVENT"] = False` since some of the code is still shared with sync and syncback pods that were not fully ported away from gevent.

- Port bin/check-attachments.py
- Port bin/inbox-auth.py
- Port bin/inbox-api.py
- Port bin/inbox-console.py
- Port bin/remove-message-attachments.py
- Port bin/restart-forgotten-accounts.py
- Port bin/syncback-stats.py

### DIFF
--- a/bin/inbox-api.py
+++ b/bin/inbox-api.py
@@ -3,6 +3,8 @@
 import os
 import sys
 
+import click
+import werkzeug.serving
 from setproctitle import setproctitle
 
 setproctitle("inbox-api")
@@ -20,15 +22,13 @@ except ImportError:
         "Try running sudo ./setup.sh"
     )
 
+inbox_config["USE_GEVENT"] = False
 
-import click
-from gevent.pywsgi import WSGIServer
 
-from inbox.api.wsgi import NylasWSGIHandler
 from inbox.error_handling import maybe_enable_rollbar
 from inbox.logging import configure_logging, get_logger
 from inbox.mailsync.frontend import SyncbackHTTPFrontend
-from inbox.util.startup import load_overrides, preflight
+from inbox.util.startup import load_overrides
 
 syncback = None
 http_server = None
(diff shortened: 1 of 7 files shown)


bin/inbox-auth.py
```diff
@@ -60,7 +59,7 @@ def main(email_address, reauth, target, provider):
 
             # Resolve unknown providers into either custom IMAP or EAS.
             if provider == "unknown":
-                is_imap = raw_input("IMAP account? [Y/n] ").strip().lower() != "n"
+                is_imap = input("IMAP account? [Y/n] ").strip().lower() != "n"
```
squeaky-pl, line 62: Opps, this is a Python 2 leftover.

bin/inbox-api.py
```diff
@@ -45,19 +45,14 @@
     default=True,
     help="Also start the syncback service",
 )
-@click.option(
-    "--enable-tracer/--disable-tracer",
-    default=True,
-    help="Disables the stuck greenlet tracer",
-)
```
squeaky-pl, line 52: Tracer is gevent specific and this code is not shared with anything else as APIs already run in threads in production, so I just removed this option.

bin/inbox-api.py
```diff
@@ -68,18 +63,23 @@ def main(prod, start_syncback, enable_tracer, config, port, enable_profiler):
...
 
-    if prod:
-        start(port, start_syncback, enable_tracer, enable_profiler)
-    else:
-        preflight()
-        from werkzeug.serving import run_with_reloader
-
-        run_with_reloader(
-            lambda: start(port, start_syncback, enable_tracer, enable_profiler)
-        )
```
squeaky-pl, line 79: The whole thing was broken because werkzeug removed `run_with_reloader` APIs at some point. It has now unified `werkzeug.serving.run_simple` API that accepts `use_reloader`.

mrhiggi-close (approved):

PR #920, opened 2024-10-02
Sync: Drop gevent

squeaky-pl:
Sync: Remove gevent

Stacked on https://github.com/closeio/sync-engine/pull/911

This again takes advantage of `IterruptibleThread` polyfill. The changes are
* changing base classes from `Greenlet` to `InterruptibleThread`.
* Added `check_interrupted` calls around I/O and other places it makes sense
* blocking calls like sleep become iterruptible etc.

### DIFF
--- a/bin/inbox-start.py
+++ b/bin/inbox-start.py
@@ -1,8 +1,5 @@
 #!/usr/bin/env python
 
-from gevent import monkey
-
-monkey.patch_all()
 
 import os
 import platform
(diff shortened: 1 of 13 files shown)


bin/inbox-start.py
```diff
@@ -149,7 +146,8 @@ def main(prod, enable_tracer, enable_profiler, config, process_num, exit_after):
     http_frontend = SyncHTTPFrontend(
         sync_service, port, enable_tracer, enable_profiler_api
     )
-    sync_service.register_pending_avgs_provider(http_frontend)
+    if enable_tracer:
+        sync_service.register_pending_avgs_provider(http_frontend)
```
squeaky-pl, line 150: This is greenlet specific, if gevent is diabled the tracer is disabled as well and it does not make sense.

inbox/mailsync/frontend.py
```diff
@@ -58,7 +58,7 @@ def profile():
         @app.route("/load")
         def load():
             if self.tracer is None:
-                return "Load tracing disabled\n", 404
+                return "Load tracing disabled\n"
```
squeaky-pl, line 61: k8s uses this for health check, the tracer is greenlet specific and disabled on threads, if we do not return 200 here the pod would never become healthy

mrhiggi-close (approved):

PR #930, opened 2024-10-07
Reduce memory pressure

squeaky-pl:
I'm currently hunting memory leaks in sync pods. While running [memray](https://github.com/bloomberg/memray) in production I noticed that we are holding a lot of references to collections of integers. This is not a surprise to me as a single folder can have tens of millions of messages these days. Nonetheless we were holding references for too long, and sometimes buffering/copying twice for no good reason. This PR fixes those problems.

image

### DIFF
--- a/inbox/crispin.py
+++ b/inbox/crispin.py
@@ -832,7 +832,10 @@ class CrispinClient:
         criteria.
 
         """
-        return sorted(int(uid) for uid in self.conn.search(criteria))
+        return sorted(
+            int(uid) if not isinstance(uid, int) else uid
+            for uid in self.conn.search(criteria)
+        )
 
     def all_uids(self) -> List[int]:
         """Fetch all UIDs associated with the currently selected folder.
@@ -881,7 +884,9 @@ class CrispinClient:
         log.debug(
             "Requested all UIDs", search_time=elapsed, total_uids=len(fetch_result)
         )
-        return sorted(int(uid) for uid in fetch_result)
+        return sorted(
+            int(uid) if not isinstance(uid, int) else uid for uid in fetch_result
+        )
 
     def uids(self, uids: List[int]) -> List[RawMessage]:
         uid_set = set(uids)
@@ -1580,4 +1585,4 @@ class GmailCrispinClient(CrispinClient):
             raise
 
         response = imapclient.response_parser.parse_message_list(data)
-        return sorted(int(uid) for uid in response)
+        return sorted(int(uid) if not isinstance(uid, int) else uid for uid in response)
--- a/inbox/mailsync/backends/imap/common.py
+++ b/inbox/mailsync/backends/imap/common.py
@@ -45,7 +45,9 @@ def local_uids(
     if limit:
         q = q.order_by(desc(ImapUid.msg_uid))
         q = q.limit(bindparam("limit"))
-    q = q.params(account_id=account_id, folder_id=folder_id, limit=limit)
+    q = q.params(account_id=account_id, folder_id=folder_id, limit=limit).yield_per(
+        1000
+    )
     return {uid for uid, in q}
 
 
--- a/inbox/mailsync/backends/imap/common.py
+++ b/inbox/mailsync/backends/imap/common.py
@@ -32,7 +32,9 @@ from inbox.models.util import reconcile_message
 log = get_logger()
 
 
-def local_uids(account_id, session, folder_id, limit=None):
+def local_uids(
+    account_id: int, session, folder_id: int, limit: "int | None" = None
+) -> "set[int]":
     q = session.query(ImapUid.msg_uid).with_hint(
         ImapUid, "FORCE INDEX (ix_imapuid_account_id_folder_id_msg_uid_desc)"
     )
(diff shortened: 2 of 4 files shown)


inbox/mailsync/backends/imap/common.py
```diff
@@ -43,8 +45,10 @@ def local_uids(account_id, session, folder_id, limit=None):
     if limit:
         q = q.order_by(desc(ImapUid.msg_uid))
         q = q.limit(bindparam("limit"))
-    results = q.params(account_id=account_id, folder_id=folder_id, limit=limit).all()
-    return {u for u, in results}
+    q = q.params(account_id=account_id, folder_id=folder_id, limit=limit).yield_per(
+        1000
+    )
```
squeaky-pl, line 50: [yield_per](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.yield_per) limits the amount of memory sqlalchemy holds at once, otherwise it will buffer everything in memory.  We buffer ourselves later to a set below. Without yield per we have two complete copies of the same data at the same time at the end of the function which increases memory pressure.

inbox/mailsync/backends/imap/generic.py
```diff
@@ -830,9 +829,15 @@ def condstore_refresh_flags(self, crispin_client):
                 interim_highestmodseq = max(v.modseq for k, v in flag_batch)
                 self.highestmodseq = interim_highestmodseq
 
+        remote_uids = crispin_client.all_uids()
+
```
squeaky-pl, line 833: Allocate this later, since we can delay this up to here.

inbox/mailsync/backends/imap/generic.py
```diff
@@ -830,9 +829,15 @@ def condstore_refresh_flags(self, crispin_client):
...
         with session_scope(self.namespace_id) as db_session:
             local_uids = common.local_uids(self.account_id, db_session, self.folder_id)
-            expunged_uids = set(local_uids).difference(remote_uids)
```
squeaky-pl, line 835: This does not need to be done while database connection is checked out.

inbox/mailsync/backends/imap/generic.py
```diff
@@ -830,9 +832,17 @@ def condstore_refresh_flags(self, crispin_client):
...
 
+        del changed_flags  # free memory as soon as possible
+
+        remote_uids = crispin_client.all_uids()
+
         with session_scope(self.namespace_id) as db_session:
             local_uids = common.local_uids(self.account_id, db_session, self.folder_id)
-            expunged_uids = set(local_uids).difference(remote_uids)
+
+        expunged_uids = local_uids.difference(remote_uids)
```
squeaky-pl, line 842: There was a `set(local_uids)` call here when this was already a `set`. This created extra copy in memory duplicating memory pressure here.

inbox/mailsync/backends/imap/generic.py
```diff
@@ -906,16 +925,19 @@ def refresh_flags_impl(self, crispin_client, max_uids):
         log.debug(
             "Changed flags refresh response, persisting changes", max_uids=max_uids
         )
-        expunged_uids = set(local_uids).difference(flags.keys())
+        expunged_uids = local_uids.difference(flags)
```
squeaky-pl, line 928: Same useless `set()` and `keys()` calls creating useless copies. In case of `keys()` dictionary are already iterated by keys so you can just pass them directly in here.

inbox/crispin.py
```diff
@@ -832,7 +832,10 @@ def search_uids(self, criteria: List[str]) -> List[int]:
         criteria.
 
         """
-        return sorted(int(uid) for uid in self.conn.search(criteria))
+        return sorted(
+            int(uid) if not isinstance(uid, int) else uid
+            for uid in self.conn.search(criteria)
```
squeaky-pl, line 837: Those are most of the time ints already (if the server follows IMAP spec), there are some nonconforming servers (minority) where we need to cast to int, but let's not do useless work on millions of them.

inbox/mailsync/backends/imap/generic.py
```diff
@@ -438,27 +444,28 @@ def initial_sync_impl(self, crispin_client):
                         self.account_id, db_session, self.folder_id
                     )
                 common.remove_deleted_uids(
-                    self.account_id,
-                    self.folder_id,
-                    set(local_uids).difference(remote_uids),
```
squeaky-pl, line 443: useless `set()` call.

inbox/mailsync/backends/imap/generic.py
```diff
@@ -438,27 +444,28 @@ def initial_sync_impl(self, crispin_client):
...
+                    self.account_id, self.folder_id, local_uids.difference(remote_uids)
                 )
 
-            new_uids = set(remote_uids).difference(local_uids)
+            new_uids = sorted(set(remote_uids).difference(local_uids), reverse=True)
```
squeaky-pl, line 450: This was sorted below later anyway, just avoid useless variable copies

mrhiggi-close (approved):

PR #936, opened 2024-10-10
Use generators to save memory

squeaky-pl:
As stated before a single folder can have millions of messages, and millions of IMAP uid integers. We were buffering to lists in intermediate functions while we can just generate one int at a time using generators and buffer to set when we really need to (because we need to calculate differences between what we have in the database and what's remotely on IMAP server to decide what to delete locally and what to create locally).

### DIFF
--- a/inbox/crispin.py
+++ b/inbox/crispin.py
@@ -362,7 +362,7 @@ uid_format = re.compile(b"[0-9]+")
 original_parse_message_list = imapclient.response_parser.parse_message_list
 
 
-def fixed_parse_message_list(data: List[bytes]) -> List[int]:
+def fixed_parse_message_list(data: List[bytes]) -> Iterable[int]:
     """Fixed version of imapclient.response_parser.parse_message_list
 
     We observed in real world that some IMAP servers send many
@@ -395,10 +395,10 @@ def fixed_parse_message_list(data: List[bytes]) -> List[int]:
     # Optimize most common textual format for low memory footprint
     with contextlib.suppress(TypeError):
         if len(data) == 1 and common_uid_list_format.match(data[0]):
-            return [
+            return (
                 int(uid_match.group(0))
                 for uid_match in re.finditer(uid_format, data[0])
-            ]
+            )
 
     return original_parse_message_list(data)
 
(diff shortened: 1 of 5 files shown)


inbox/crispin.py
```diff
@@ -825,19 +826,19 @@ def condstore_supported(self) -> bool:
...
 
-    def search_uids(self, criteria: List[str]) -> List[int]:
+    def search_uids(self, criteria: List[str]) -> Iterable[int]:
         """
         Find UIDs in this folder matching the criteria. See
         http://tools.ietf.org/html/rfc3501.html#section-6.4.4 for valid
         criteria.
 
         """
-        return sorted(
```
squeaky-pl, line 835: The sorted call here was doing useless extra work (creating a sorted list) and adding memory pressure, because we are converting most of the time to sets in the caller. When we are *not* converting to sets the order does not matter for the caller.

inbox/crispin.py
```diff
@@ -881,12 +882,8 @@ def all_uids(self) -> List[int]:
                 raise
 
         elapsed = time.time() - t
-        log.debug(
-            "Requested all UIDs", search_time=elapsed, total_uids=len(fetch_result)
-        )
-        return sorted(
```
squeaky-pl, line 887: again useless sorted call like above.

inbox/crispin.py
```diff
@@ -881,12 +882,8 @@ def all_uids(self) -> List[int]:
                 raise
 
         elapsed = time.time() - t
-        log.debug(
-            "Requested all UIDs", search_time=elapsed, total_uids=len(fetch_result)
-        )
```
squeaky-pl, line 886: fetch_result is now a generator, if we wanted to know the length here we would need to convert it to list - which would use much more memory. Anyway it's not really interesting to know the size here.

mrhiggi-close (approved):
That is a lot of sorting removed too, nice! Was this found by reviewing the memory profile? I'd like to see how you're gathering that/looking for.

squeaky-pl:
Yes, I found this using memray. I will do a S&T because this journey although much longer than I expected is teaching me a lot.

PR #937, opened 2024-10-13
Kill greenlets faster

squeaky-pl:
This is a bugfix for https://github.com/closeio/sync-engine/pull/921.

When the original PR was deployed I noticed that the sync shutdown has become much slower in production i.e. minutes instead of seconds. This made the whole deploy process much slower. This is because I was now correctly waiting for all the greenlets but I was killing them one by one mostly and waiting for them to finish one by one at the same time. I know tweaked this to kill them one by one without waiting and then waiting for all of them to finish.

There's also a bug that crept in again because of too many branches/too much rebasing.

### DIFF
--- a/bin/inbox-start.py
+++ b/bin/inbox-start.py
@@ -142,8 +142,8 @@ def main(prod, enable_tracer, enable_profiler, config, process_num, exit_after):
 
     sync_service = SyncService(process_identifier, process_num)
 
-    signal.signal(signal.SIGTERM, sync_service.stop)
-    signal.signal(signal.SIGINT, sync_service.stop)
+    signal.signal(signal.SIGTERM, lambda *_: sync_service.stop())
+    signal.signal(signal.SIGINT, lambda *_: sync_service.stop())
     prepare_exit_after(log, sync_service, exit_after)
 
     http_frontend = SyncHTTPFrontend(
--- a/inbox/mailsync/service.py
+++ b/inbox/mailsync/service.py
@@ -4,7 +4,6 @@ import time
 from threading import BoundedSemaphore
 from typing import Type
 
-import gevent
 from sqlalchemy import and_, or_
 from sqlalchemy.exc import OperationalError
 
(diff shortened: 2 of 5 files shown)


bin/inbox-start.py
```diff
@@ -142,8 +142,8 @@ def main(prod, enable_tracer, enable_profiler, config, process_num, exit_after):
 
     sync_service = SyncService(process_identifier, process_num)
 
-    signal.signal(signal.SIGTERM, sync_service.stop)
-    signal.signal(signal.SIGINT, sync_service.stop)
+    signal.signal(signal.SIGTERM, lambda *_: sync_service.stop())
+    signal.signal(signal.SIGINT, lambda *_: sync_service.stop())
```
squeaky-pl, line 146: I made this mistake too many times b/c too much branches. `stop()` now does not accept arguments and signal handlers do.

inbox/mailsync/service.py
```diff
@@ -118,24 +117,24 @@ def run(self):
...
 
+        kill_all(self.contact_sync_monitors.values())
+        self.log.info(
+            "stopped contact sync monitors", count=len(self.contact_sync_monitors)
+        )
+        kill_all(self.event_sync_monitors.values())
+        self.log.info(
+            "stopped event sync monitors", count=len(self.event_sync_monitors)
+        )
+
```
squeaky-pl, line 128: `event_sync_monitors` and `contact_sync_monitors` are easier/faster to kill because they are just single greenlets as opposed to trees of greenlets in `email_sync_monitors`. It makes sense to kill them first because of that. The order is not really meaningful from correctness point of view.

inbox/mailsync/service.py
```diff
@@ -118,24 +117,24 @@ def run(self):
...
         for email_sync_monitor in self.email_sync_monitors.values():
             if email_sync_monitor.delete_handler:
                 email_sync_monitor.delete_handler.kill()
+            kill_all(email_sync_monitor.folder_monitors, block=False)
```
squeaky-pl, line 132: `folder_monitors` are children of `email_sync_monitor` and `email_sync_monitor` kills them in the cleanup but it makes sense to start killing them as soon as possible without blocking as it speeds up the process shutdown process. Again order here is not really meaningful from the correctness point of views.

drewler (approved):

PR #940, opened 2024-10-14
Stop mail syncs in parallel

squeaky-pl:
To stop even faster (I care about this because it affects how fast deploys go out) we can stop all the email syncs in parallel since they are totally independent and order does not matter.

### DIFF
--- a/inbox/mailsync/backends/imap/monitor.py
+++ b/inbox/mailsync/backends/imap/monitor.py
@@ -11,6 +11,7 @@ from inbox.mailsync.gc import DeleteHandler
 from inbox.models import Account, Folder
 from inbox.models.category import Category, sanitize_name
 from inbox.models.session import session_scope
+from inbox.util.concurrency import kill_all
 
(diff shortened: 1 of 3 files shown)


inbox/mailsync/backends/imap/monitor.py
```diff
@@ -185,3 +186,14 @@ def sync(self):
...
+    def stop(self) -> None:
+        from inbox.mailsync.backends.gmail import GmailSyncMonitor
+
+        if self.delete_handler:
+            self.delete_handler.kill()
+        kill_all(self.folder_monitors, block=False)
+        if isinstance(self, GmailSyncMonitor):
+            kill_all(self.label_rename_handlers.values(), block=False)
+        self.sync_greenlet.kill(block=False)
+        self.join()
```
squeaky-pl, line 199: This is better contained in a function since you need to stop the whole "tree" of greenlets.

drewler (approved):

PR #941, opened 2024-10-14
Only free inbox_uids if defined

mrhiggi-close:
Testing email syncing, I encountered an error.

When the `if self.is_all_mail(crispin_client):` path isn't taken (because email can't sync?) then there is no `inbox_uids` defined.
```
  File "/opt/app/inbox/util/concurrency.py", line 75, in wrapped
    return func(*args, **kwargs)
  File "/opt/app/inbox/mailsync/backends/imap/generic.py", line 291, in _run_impl
    self.state = self.state_handlers[old_state]()
  File "/opt/app/inbox/util/concurrency.py", line 75, in wrapped
    return func(*args, **kwargs)
  File "/opt/app/inbox/mailsync/backends/imap/generic.py", line 421, in initial_sync
    self.initial_sync_impl(crispin_client)
  File "/opt/app/inbox/mailsync/backends/gmail.py", line 300, in initial_sync_impl
    del inbox_uids  # free up memory as soon as possible
UnboundLocalError: local variable 'inbox_uids' referenced before assignment
```

### DIFF
--- a/inbox/mailsync/backends/gmail.py
+++ b/inbox/mailsync/backends/gmail.py
@@ -293,11 +293,11 @@ class GmailFolderSyncEngine(FolderSyncEngine):
                 uids_to_download = sorted(unknown_uids - inbox_uids) + sorted(
                     unknown_uids & inbox_uids
                 )
+                del inbox_uids  # free up memory as soon as possible
             else:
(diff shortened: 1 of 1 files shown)


wojcikstefan (approved):

squeaky-pl:
Oops

PR #943, opened 2024-10-15
Update `HIGHESTMODSEQ` cache when decremented by server

neob91-close:
Fixes #

Summary: Some servers can decrement the `HIGHESTMODSEQ` value while the `UIDVALIDITY` remains unchanged, which is against the spec. This results in syncing issues.

Test Plan:

Reviewers:


squeaky-pl (approved):

squeaky-pl:
This needs a rebase now, if you want to keep trying this on a single pod in production.

PR #944, opened 2024-10-15
Interruptible threading

squeaky-pl:
This is a secret sauce that will let me convert syncback and sync pods to threading without doing a complete rewrite. Both sync and syncback processes are trees of greenlets and they need to become trees of threads.

Greenlets can be killed in any place you are calling gevent-native API or monkey patched stdlib. Internally this raises exceptions both in greenlets and on this branch but normally you don't see this because it happens behind the scenes. I provided a polyfill that lets me replace the Greenlet base classes with InterruptibleThread. This polyfill provides greenlet-like methods/properties and emulates greenlets as close as practically possible. The threads must call `check_interrupted` from time to time or use conveniently provided `interruptible_threading.sleep` and others utilities that will do that for you.

The polyfill is less than 100 lines of Python code if you remove docstrings and makes porting straghtforward.

For this in action please see: https://github.com/closeio/sync-engine/pull/911

### DIFF
--- a/inbox/interruptible_threading.py
+++ b/inbox/interruptible_threading.py
@@ -70,7 +70,7 @@ class InterruptibleThread(threading.Thread):
         thread is started. Otherwise, the subclass must implement the `_run`
         method.
         """
-        self.__should_be_killed = False
+        self.__should_be_killed = threading.Event()
         self.__ready = False
         self.__run_target = (
             _InterruptibleThreadTarget(target, args, kwargs) if target else None
@@ -130,7 +130,7 @@ class InterruptibleThread(threading.Thread):
 
         If block is True, wait until the thread is ready.
         """
-        self.__should_be_killed = True
+        self.__should_be_killed.set()
         if block:
             self.join()
 
@@ -141,7 +141,7 @@ class InterruptibleThread(threading.Thread):
         Don't use this directly instead use the public
         `check_interrupted` function below.
         """
-        if self.__should_be_killed:
+        if self.__should_be_killed.wait(0.01):
             raise InterruptibleThreadExit()
 
         if (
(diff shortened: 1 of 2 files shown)


inbox/interruptible_threading.py
```diff
@@ -0,0 +1,263 @@
...
+        except queue.Empty:
+            current_thread._check_interrupted()
+
+
+@_interruptible(lambda: None)
+def check_interrupted(current_thread: InterruptibleThread, /) -> None:
+    """
+    Check if the current thread is interrupted.
+    """
+    return current_thread._check_interrupted()
```
mrhiggi-close, line 226: Why return a method call that returns `None` and only communicates by raising an exception?

squeaky-pl: This is exactly how gevent works.

squeaky-pl: This is what https://greenlet.readthedocs.io/en/latest/api.html#greenlet.GreenletExit does. In gevent if you call `.kill()` on a greenlet it will raise `GreenletExit` next time it uses one of the gevent APIs or monkeypatched stdlib.

mrhiggi-close (approved):
Nothing here that needs to hold the merge

inbox/interruptible_threading.py
squeaky-pl: @mrhiggi-close Also asyncio works exactly the same way in this regard, it raises an exception https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError if you `.cancel()` a running task.

squeaky-pl: Control flow with exceptions is a standard thing with concurrency libraries. I will cover this in S&T.

PR #948, opened 2024-10-18
Lock some CPU & RAM intesitve operations

squeaky-pl:
We are running on threads in production now. I observed CPU & RAM usage balloon on some of the pods where we run a lot of accounts. This is because with gevent CPU & RAM intensive sections are "naturally" locked - i.e. gevent will only switch between greenlets when you use gevent APIs or I/O - threads on the other hand can switch every N-th Python bytecode. This in turn makes it possible to allocate faster and have bigger spikes in CPU usage. I tried this branch already in production and it solves the problem without slowing down or negatively impacting syncs. It just brings back the "natural" gevent behavior.

This adds `with self.global_lock` sections around CPU & RAM intensive operations so threads don't switch between each other in those sections and don't inflate CPU & RAM usage. The diff is ugly because git can't understand indent - dedent.

We might need to lock more sections in the future, but this is enough for now not to trigger warnings/alarms.

### DIFF
--- a/inbox/mailsync/backends/imap/generic.py
+++ b/inbox/mailsync/backends/imap/generic.py
@@ -63,6 +63,7 @@ sessions reduce scalability.
 
 import contextlib
 import imaplib
+import threading
 import time
 from datetime import datetime, timedelta
 from typing import Any, Dict, NoReturn, Optional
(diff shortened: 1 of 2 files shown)


inbox/mailsync/backends/imap/generic.py
```diff
@@ -154,6 +155,8 @@ def __repr__(self) -> str:
 class FolderSyncEngine(InterruptibleThread):
     """Base class for a per-folder IMAP sync engine."""
 
+    global_lock = threading.BoundedSemaphore(1)
```
squeaky-pl, line 158: It's not just a lock but a BoundedSemaphore because I might consider increasing it later to 2 or something else. Also sync-engine uses BoundedSemaphore everywhere else and I wanted to stay consistent.

drewler (approved):

PR #953, opened 2024-10-22
Goodbye Gevent

squeaky-pl:
We've been running without Gevent in production since last Thursday. It's time to cleanup the de-fact dead code.

* Removes `USE_GEVENT` config option and the associated gevent waiter for MySQL
* Removes GreenletTracer and KillerGreenletTracer - those were greenlet-specific and responsible for monitoring and killing stuck greenlets that got stuck on CPU intensive parts, threads can switch by themselves every N-th instruction without performing I/O
* Remove gevent and it's dependencies from the requirements
* Removes mentions of gevent and greenlets from the code


bin/inbox-start.py
```diff
@@ -58,11 +58,6 @@
     default=False,
     help="Disables the autoreloader and potentially other non-production features.",
 )
-@click.option(
-    "--enable-tracer/--disable-tracer",
-    default=True,
-    help="Disables the stuck greenlet tracer",
-)
```
squeaky-pl, line 65: The tracers are `GreenletTracer` and `KillerGreenletTracer` which are gevent-specific and it does not make sense to have this option anymore.

inbox/mailsync/frontend.py
```diff
@@ -57,12 +45,7 @@ def profile():
 
         @app.route("/load")
         def load():
-            if self.tracer is None:
-                return "Load tracing disabled\n"
-            resp = jsonify(self.tracer.stats())
-            if request.args.get("reset ") in (1, "true"):
-                self.tracer.reset()
-            return resp
+            return "Load tracing disabled\n"
```
squeaky-pl, line 48: I cannot remove this as k8s uses this as a healthcheck.

requirements/prod.txt
```diff
@@ -85,6 +84,4 @@ wcwidth==0.2.6
 WebOb==1.8.8
 Werkzeug==3.0.3
 zipp==3.20.2
-zope.event==4.5.0
-zope.interface==5.4.0
```
squeaky-pl, line 89: Dependencies of [gevent](https://github.com/gevent/gevent/blob/master/setup.py#L231-L238) .

drewler (approved):
...wow, awesome! ✨🧹

PR #956, opened 2024-10-24
Use threading.Event() in "interruptible sleep"

squeaky-pl:
Still in the world of CFS throttling...

TLDR: When I ported from gevent to threading I implemented a "polyfill" for threads that implements the APIs very similar to greenlets (or at least the subset sync-engine uses). If you wanna read more on "interruptible threading" check out this [PR](https://github.com/closeio/sync-engine/pull/944). But if you don't feel like - there's gonna be a show & tell ;-)

Threads can not be killed or interrupted (in Python, can be generally in C with POSIX threads)  by another thread unless they cooperatively check from time to time if they are supposed to exit and then kindly exit.

Greenlets on the other hand can be interrupted e.g. by timeouts and can be also `kill()` ed. In gevent world every time you yield back to the loop (either by using gevent APIs or monkeypatched stdlib) this action is handled. Sync-engine uses timeouts and uses (abuses?) `kill()`  extensively so I had to come up with "interruptible sleeps" and "interruptible threading".

I initially implemented this with a semi [busy-waiting](https://en.wikipedia.org/wiki/Busy_waiting), i.e. sleeping 10 seconds means we sleep for 200 ms and then wake up to check if we are supposed to be killed or we timeout and then sleep again until we reach 10 seconds. This had bigger impact on CFS throttling (i.e. how much k8s has to pause pods to keep them under CPU limits) than I envisioned. Probably because that means we switch between OS threads a lot (every 200ms) and that's not free (as in we go through user-space to kernel-space and user-space again a lot).

This can be also implemented with a `threading.Event()`, this event is initially unset and we just `.wait()` for it until the sleep elapses. If thread is `kill()` ed we will set the `threading.Event()` and if it happens to be sleeping it will wake up immediately before the sleep elapses.

Sync-engine sync pod sleep a lot. Let's give them a good sleep. After this change CFS throttling is barely visible in the metrics (left before, right after).

Screenshot 2024-10-24 at 14 42 41
Screenshot 2024-10-24 at 14 42 52
Screenshot 2024-10-24 at 14 43 12

There are tests for those things already in https://github.com/closeio/sync-engine/blob/6f13f79d920a80e73c76006fcd4e31f0564062e5/tests/test_interruptible_threading.py#L48-L61 and  https://github.com/closeio/sync-engine/blob/6f13f79d920a80e73c76006fcd4e31f0564062e5/tests/test_interruptible_threading.py#L96-L122.

### DIFF
--- a/inbox/interruptible_threading.py
+++ b/inbox/interruptible_threading.py
@@ -70,6 +70,7 @@ class InterruptibleThread(threading.Thread):
         thread is started. Otherwise, the subclass must implement the `_run`
         method.
         """
+        self._should_be_interrupted = threading.Event()
         self.__should_be_killed = False
         self.__ready = False
         self.__run_target = (
@@ -131,6 +132,7 @@ class InterruptibleThread(threading.Thread):
         If block is True, wait until the thread is ready.
         """
         self.__should_be_killed = True
+        self._should_be_interrupted.set()
         if block:
             self.join()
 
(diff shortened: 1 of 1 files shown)


inbox/interruptible_threading.py
```diff
@@ -70,6 +70,7 @@ def __init__(
         thread is started. Otherwise, the subclass must implement the `_run`
         method.
         """
+        self._should_be_interrupted = threading.Event()
```
drewler, line 73: nit: shouldn't we use double underscores for naming consistency?

drewler (approved):

inbox/interruptible_threading.py
squeaky-pl: It's because it's not really private to the class, the free functions are using it.

PR #957, opened 2024-10-24
Skip SQLAlchemy when fetching millions of UIDs

squeaky-pl:
This function is called a lot and it shows in the profiles as the busiest function. I've used [py-spy](https://github.com/benfred/py-spy) to obtain those profiles in production.

I discovered that the SQLAlchemy "post-processing" which just creates intermediate objects to handle all the possible shapes that SQLAlchemy supports takes significant portion of this function runtime. You can see a lot of SQLAlchemy fiddling below the call to it, the actual execution in MySQL is 50% of time.

Screenshot 2024-10-24 at 19 13 42

Since this is a trivial query that returns a single column we can easily rewrite it to use driver's DB-API which is still portable between different databases ([PEP](https://peps.python.org/pep-0249/)) which means that it's 2 times faster now as it only spends time executing.

Screenshot 2024-10-24 at 19 11 06

Just the other day I was dealing with a case when this function can return 24 million rows for a single folder, so performance here is critical.

### DIFF
--- a/inbox/mailsync/backends/imap/common.py
+++ b/inbox/mailsync/backends/imap/common.py
@@ -28,6 +28,7 @@ from inbox.models.backends.imap import ImapFolderInfo, ImapUid
 from inbox.models.category import Category
 from inbox.models.session import session_scope
 from inbox.models.util import reconcile_message
+from inbox.sqlalchemy_ext.util import get_db_api_cursor_with_query
 
 log = get_logger()
 
@@ -45,10 +46,11 @@ def local_uids(
     if limit:
         q = q.order_by(desc(ImapUid.msg_uid))
         q = q.limit(bindparam("limit"))
-    q = q.params(account_id=account_id, folder_id=folder_id, limit=limit).yield_per(
-        1000
-    )
-    return {uid for uid, in q}
+    q = q.params(account_id=account_id, folder_id=folder_id, limit=limit)
+
+    db_api_cursor = get_db_api_cursor_with_query(session, q)
+
+    return {uid for uid, in db_api_cursor.fetchall()}
 
 
 def lastseenuid(account_id, session, folder_id):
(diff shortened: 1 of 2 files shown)


inbox/mailsync/backends/imap/common.py
```diff
@@ -45,10 +53,12 @@ def local_uids(
     if limit:
         q = q.order_by(desc(ImapUid.msg_uid))
         q = q.limit(bindparam("limit"))
-    q = q.params(account_id=account_id, folder_id=folder_id, limit=limit).yield_per(
-        1000
-    )
```
squeaky-pl, line 50: yield per is SQLAlchemy specific to save on RAM with all those extra intermediate objects that SQLAlchemy is creating and is no longer relevant as we skip them.

inbox/sqlalchemy_ext/util.py
```diff
@@ -361,3 +362,24 @@ def safer_yield_per(query, id_field, start_id, count):
...
+    """
+    Return a DB-API cursor with the given SQLAlchemy query executed.
+
+    This is useful when you want to optimize and skip the machinery of
+    SQLAlchemy ORM, particularly when you want to execute
+    a query that returns a large number of rows with a couple of columns.
+    SQLAlchemy ORM has to instantiate several Python objects for each row
+    returned by the query, which can be a performance bottleneck.
+    """
+    compiled_query = query.statement.compile(dialect=mysql_dialect)
```
drewler, line 380: Could we use something like `session.get_bind().dialect` here to avoid hardcoding `mysql_dialect`?

drewler (approved):
🚀

inbox/sqlalchemy_ext/util.py
squeaky-pl: Done in [`d8fac30` (#957)](https://github.com/closeio/sync-engine/pull/957/commits/d8fac3081e9defd222b3befd39dbc39e59c3366e)

PR #439, opened 2022-12-09
Python 3.10

squeaky-pl:

PR #959, opened 2024-10-28
Python 3.10

squeaky-pl:
Supersedes: https://github.com/closeio/sync-engine/pull/439

### DIFF
--- a/inbox/util/html.py
+++ b/inbox/util/html.py
@@ -68,7 +68,14 @@ def strip_tags(html: str) -> str:
     You are responsible for handling it in the calling function.
     """
     s = HTMLTagStripper()
-    s.feed(html)
+
+    try:
+        s.feed(html)
+    except AssertionError as e:
+        if e.args[0].startswith("unknown status keyword"):
+            raise HTMLParseError(*e.args) from e
+        raise
+
     return s.get_data()
 
 
--- a/inbox/util/html.py
+++ b/inbox/util/html.py
@@ -9,7 +9,7 @@ else:
     from html import escape as html_escape
 
 
-class HTMLParseError(Exception):
+class HTMLParseError(ValueError):
     pass
 
 
@@ -42,11 +42,6 @@ class HTMLTagStripper(HTMLParser):
         if not self.strip_tag_contents_mode:
             self.fed.append(d)
 
-    if (3,) <= sys.version_info < (3, 10):
-
-        def error(self, message):
-            raise HTMLParseError(message)
-
     def handle_entityref(self, d):
         try:
             val = chr(name2codepoint[d])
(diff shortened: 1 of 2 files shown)


inbox/util/html.py
```diff
@@ -68,7 +68,14 @@ def strip_tags(html: str) -> str:
...
     """
     s = HTMLTagStripper()
-    s.feed(html)
+
+    try:
+        s.feed(html)
+    except AssertionError as e:
+        if e.args[0].startswith("unknown status keyword"):
+            raise HTMLParseError(*e.args) from e
+        raise
```
wojcikstefan, line 77: Curious, why was this change needed?

squeaky-pl: The python parser in Python 3.10 started throwing AssertionErrors on some invalid HTML, there were previously HTMLParserError. I think it's opinionated weather it should be AssertionErrors, especially when treating with user supplied data (which emails are). We actually have a test for those, meaning that at some point Nylas people hit this kind of HTML that would now trigger AssertionError.

squeaky-pl: @wojcikstefan should I add clarification comment maybe?

wojcikstefan: Thanks for explaining, makes sense 👍 Yea, I think it's worth adding a comment – it might not be clear at a glance why we don't work directly with the `AssertionError` further up the stack.

Are there other invalid HTML errors that will still raise an `AssertionError`? Should we re-raise **all** `s.feed(html)` assertion errors as HTML parser errors? Or does `s.feed` do more than just parse?

squeaky-pl: Additionally if those would become AssertionErrors we would not catch them [here](https://github.com/closeio/sync-engine/blob/4f192bca8af9c8f1b91e25a0809cf14954a3d64c/inbox/models/message.py#L656-L665)

```
    def calculate_html_snippet(self, text: str) -> str:
        try:
            text = strip_tags(text)
        except HTMLParseError:
            log.error(
                "error stripping tags", message_nylas_uid=self.nylas_uid, exc_info=True
            )
            text = ""

        return self.calculate_plaintext_snippet(text)
```

meaning that we would discard the entire email, instead of discarding just the snippet.

wojcikstefan: We technically could replace `except HTMLParserError` with `except AssertionError` though, right? Or are there other `AssertionErrors` being thrown that we should definitely NOT catch?

squeaky-pl: @wojcikstefan I did a deep dive, found the change in CPython that causes this:

https://github.com/python/cpython/commit/e34bbfd61f405eef89e8aa50672b0b25022de320

As you can see they have removed the [error callback](https://github.com/python/cpython/commit/e34bbfd61f405eef89e8aa50672b0b25022de320#diff-7304624ffcc8be1268e5f86ebf1723034477694ce33456dc330f9c03f2d1e84fL32-L35), we were using the error callback to raise [our exception](https://github.com/closeio/sync-engine/blob/3fc0382512bba63e1a9fe551910c2a069b20615a/inbox/util/html.py#L46-L48). But that method was [long deprecated](https://github.com/python/cpython/issues/76025). I still think the choice to raise AssertionError is the wrong choice of CPython developers. So I agree we should catch all the AssertionErrors.

There are some AssertionErrors that are programming errors in the source of the parser, like [this one](https://github.com/python/cpython/blob/ffee63f34445412322a7afc2535731be221cabba/Lib/html/parser.py#L382C7-L382C56), but they should not happen.

Thanks for catching this.  Will follow with a fix.

squeaky-pl: @wojcikstefan Pushed a fix in [`4ae8c49` (#959)](https://github.com/closeio/sync-engine/pull/959/commits/4ae8c4975a445914b2ebc79587b14c351356eee5). Assigning you as a reviewer as well because you spent a lot of time looking at this already.

wojcikstefan (approved):
Looks good! Do you think it's worth rolling out partially first to confirm we didn't miss anything?

squeaky-pl:
@wojcikstefan Yes, I will definitely canary this.

PR #960, opened 2024-10-28
Python 3.12

squeaky-pl:
I've already started testing this end to end on single pods where my email is synced, will of course do a canary.

# Based on PR #959

TODO:

* [x] Replace boto with boto3
* [x] Open [issue](https://github.com/closeio/limitlion/issues/37) on limitlion about setuptools dependency on Python 3.12

## Full chain of PRs as of 2024-10-28

* PR #960:
  `python-312` ➔ `python-310`
* PR #959:
  `python-310` ➔ `master`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # --- Stage 0 --- #
 # This first stage is responsible for installing any dependencies the app needs
 # to run, and updating any base dependencies.
-FROM ubuntu:22.04 AS stage_0
+FROM ubuntu:24.04 AS stage_0
 
 RUN groupadd -g 5000 sync-engine \
   && useradd -d /home/sync-engine -m -u 5000 -g 5000 sync-engine
@@ -24,7 +24,7 @@ RUN echo $BUILD_WEEK && apt-get update \
     gpg \
     gpg-agent \
     dirmngr \
-    python3.10 \
+    python3.12 \
     gettext-base \
     libmysqlclient21 \
     mysql-client \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -54,16 +54,16 @@ RUN apt-get update \
...
     pkg-config \
-    python3.9-dev \
+    python3.12-dev \
+    python3.12-venv \
     python3-pip \
     libmysqlclient-dev \
   && rm -rf /var/lib/apt/lists/*
 
 COPY /requirements/ /requirements/
-RUN python3.9 -m pip install pip==24.0 virtualenv==20.25.1 && \
```
squeaky-pl, line 63: Python in Ubuntu 24.04 started complaining if you install packages into the operating system using pip, because you are supposed to install them using apt-get, and use virtualenv for everything else. I was doing that before because we had terribly outdated pip otherwise, but now we caught up with Python 3.12 so we can just use builtin venv from apt-get to create the virtualenv.

requirements/prod.txt
```diff
@@ -69,6 +69,7 @@ requests-file==1.5.1
 rollbar==0.16.2
 s3transfer==0.10.3
 setproctitle==1.2.2
+setuptools==75.2.0
```
squeaky-pl, line 72: See: https://github.com/closeio/limitlion/issues/37

drewler (approved):
🐍🚀

drewler (approved):

wojcikstefan:
Seems like we missed a spot: https://github.com/closeio/sync-engine/blob/c6fbe9913de564aa99a66a457e7a7bfb95b590c7/.python-version

PR #961, opened 2024-10-29
Use boto3

squeaky-pl:
The code that is uploading/downloading things from S3 still used outdated [boto](https://pypi.org/project/boto/) library (last released in 2018 💀). That library is no longer maintained and does not work on newer Python versions. I dropped that old library and ported it to boto3 (which we had in requirements but were not using at all 🤡). I updated boto3 at the same time.

Note that we are using minio locally to do real requests to S3 test double, so this is fairly well tested but I'll canary as well.

### DIFF
--- a/inbox/util/blockstore.py
+++ b/inbox/util/blockstore.py
@@ -1,8 +1,9 @@
+import io
 import os
 import time
 from collections.abc import Callable
 from hashlib import sha256
-from typing import Optional
+from typing import Iterable, Optional
 
 import zstandard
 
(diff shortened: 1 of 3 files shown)


inbox/util/blockstore.py
```diff
@@ -313,7 +332,9 @@ def _delete_from_s3_bucket(data_sha256_hashes, bucket_name):
     # Boto pools connections at the class level
     bucket = get_s3_bucket(bucket_name)
 
-    bucket.delete_keys([key for key in data_sha256_hashes], quiet=True)
+    bucket.delete_objects(
+        Delete={"Objects": [{"Key": key} for key in data_sha256_hashes], "Quiet": True}
+    )
```
squeaky-pl, line 337: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/bucket/delete_objects.html

inbox/util/blockstore.py
```diff
@@ -147,14 +147,29 @@ def _save_to_s3(data_sha256: str, data: bytes, *, overwrite: bool = False) -> No
...
+    """
+    Check if a key exists in an S3 bucket by doing a HEAD request.
+    """
+    try:
+        bucket.Object(key).load()
+    except botocore.exceptions.ClientError as e:
+        if e.response["Error"]["Code"] == "404":
+            return False
+        else:
+            raise
```
squeaky-pl, line 170: So annoying this became hard in boto3: https://stackoverflow.com/a/33843019

drewler (approved):

PR #967, opened 2024-11-04
Bugfix: Really delete message hashes

squeaky-pl:


inbox/mailsync/gc.py
```diff
@@ -164,6 +170,8 @@ def check(self, current_time):
                 # transaction.
                 db_session.commit()
 
+        delete_message_hashes(self.namespace_id, self.account_id, dangling_sha256s)
```
drewler, line 173: Does this mean we'll `delete_message_hashes` as part of `check`? I guess it makes sense since we also delete them from the DB here, just wonder if performing the actual delete from here could have any implications (e.g. related to performance).

Separatedly: does it need `dry_run=True`?

squeaky-pl: Of course it can slow it down slightly but:
- There are separate DeleteHandler for every account and they all run in parallel
- In practice deleting 1000 S3 objects takes a couple of milliseconds and can be done with one request to S3

So I don't think in practically matters, but I will monitor of course.

drewler (approved):

PR #968, opened 2024-11-05
sync-single-account.py

squeaky-pl:
This is something that I did over and over using sync-engine console to debug escalations and otherwise troubleshoot syncing of a single account. Since it is not obvious how to do it I wrapped it in a script so everybody can use it.

### DIFF
--- a/inbox/mailsync/backends/base.py
+++ b/inbox/mailsync/backends/base.py
@@ -1,4 +1,5 @@
 import threading
+from typing import TYPE_CHECKING, ClassVar
 
 from inbox.config import config
 from inbox.interruptible_threading import InterruptibleThread, InterruptibleThreadExit
@@ -6,6 +7,9 @@ from inbox.logging import get_logger
 from inbox.models.session import session_scope
 from inbox.util.concurrency import kill_all, retry_with_logging
 from inbox.util.debug import bind_context
+
+if TYPE_CHECKING:
+    from inbox.mailsync.backends.imap.generic import FolderSyncEngine
 
 log = get_logger()
 
@@ -22,6 +26,8 @@ class MailsyncDone(InterruptibleThreadExit):
 
 
 class BaseMailSyncMonitor(InterruptibleThread):
+    sync_engine_class: ClassVar[type["FolderSyncEngine"]]
+
     """
     The SYNC_MONITOR_CLS for all mail sync providers should subclass this.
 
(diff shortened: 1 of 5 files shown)


inbox/mailsync/backends/base.py
```diff
@@ -22,6 +26,8 @@ class MailsyncDone(InterruptibleThreadExit):
 
 
 class BaseMailSyncMonitor(InterruptibleThread):
+    sync_engine_class: ClassVar[type["FolderSyncEngine"]]
+
```
squeaky-pl, line 30: I've moved this out of an initializer so the thread class used to sync a single folder can be accessed without constructing a monitor. A monitor is a parent thread used to sync all the folders.

inbox/mailsync/backends/gmail.py
```diff
@@ -55,174 +55,6 @@
 MAX_DOWNLOAD_COUNT = 1
 
 
-class GmailSyncMonitor(ImapSyncMonitor):
```
squeaky-pl, line 58: I just switched the order of classes here so I can access `GmailFolderSyncEngine` from `GmailSyncMonitor`

inbox/mailsync/service.py
```diff
@@ -48,6 +50,24 @@ def shared_sync_event_queue_for_zone(zone):
...
+        module.PROVIDER: getattr(module, module.SYNC_MONITOR_CLS)
+        for module in module_registry.values()
+        if hasattr(module, "SYNC_MONITOR_CLS")
+    }
+
+    for provider_name, _ in providers.items():
+        if provider_name not in monitor_classes:
+            monitor_classes[provider_name] = monitor_classes["generic"]
+
+    return monitor_classes
```
squeaky-pl, line 68: This was extracted from `SyncService` intializer so I can use this from the script. This code is itself terribly dynamic but sync-engine was built to be extended by installing external modules like plugins, something that we don't do ourselves but I don't want to change it right n ow.

inbox/mailsync/backends/gmail.py
```diff
@@ -598,3 +430,172 @@ def g_msgids(namespace_id, session, in_):
         .all()
     )
     return {g_msgid for g_msgid, in query}
+
+
+class GmailSyncMonitor(ImapSyncMonitor):
+    sync_engine_class: ClassVar[type[FolderSyncEngine]] = GmailFolderSyncEngine
```
squeaky-pl, line 436: And this code was moved below with

`sync_engine_class: ClassVar[type[FolderSyncEngine]] = GmailFolderSyncEngine`

moved from the initializer to class body.

neob91-close (approved):

PR #969, opened 2024-11-06
Limit ORM usage for ImapUids objects

squeaky-pl:
We have some non-conforming generic IMAP servers that sometimes can show the same message under thousands of IMAP uids.

Normally a message can show at most two, maybe three times if it's moved between different folders quickly as there's inevitably a delay before we detect it disappeared and sometimes it will be after we already detected addition to another folder. There's also not a standard way of detecting moves in IMAP so it looks like additions and deletions to sync-engine so this is just how things are.

But sometimes there are those generic IMAP servers that report the same message appearing thousands of times in the same folder, which is nonsense, and sync-engine was not ready for that as it tried to load all the ImapUids objects using ORM which exploded CPU and RAM usage. I've rewritten the parts that care about the count of ImapUids using just `exists` queries, and put a sanity limits when it actually has to load them in memory to figure out how categorize messages.

The main problem I am trying to solve here is this was affecting other accounts running in the same pod, leading to OOM kills and slowing other accounts, affecting their sync experience. It's what I call "the noisy neighbor problem".

I've already tested this on one of such accounts and it improves the situation... substantially:

Screenshot 2024-11-06 at 11 58 09

### DIFF
--- a/inbox/mailsync/backends/imap/common.py
+++ b/inbox/mailsync/backends/imap/common.py
@@ -262,15 +262,13 @@ def remove_deleted_uids(account_id, folder_id, uids):
             db_session.delete(imapuid)
 
             if message is not None:
-                message_imapuids_count = (
+                message_imapuids_exist = db_session.query(
                     imapuids_for_message_query(
                         ImapUid, account_id=account_id, message_id=message.id
-                    )
-                    .with_session(db_session)
-                    .count()
-                )
+                    ).exists()
+                ).scalar()
 
-                if not message_imapuids_count and message.is_draft:
+                if not message_imapuids_exist and message.is_draft:
                     # Synchronously delete drafts.
                     thread = message.thread
                     if thread is not None:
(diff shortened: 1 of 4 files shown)


inbox/mailsync/gc.py
```diff
@@ -59,15 +56,13 @@ def __init__(
         account_id,
         namespace_id,
         provider_name,
-        uid_accessor,
```
squeaky-pl, line 62: This does not need to be passed into the initializer as we were passing it the same thing everywhere.

inbox/mailsync/backends/imap/common.py
```diff
@@ -101,9 +112,9 @@ def update_message_metadata(
         # (and in turn one category) depending on the order they were returned
         # from the database. This makes it deterministic and more-correct because a message
         # is likely in a folder (and category) it was added to last.
-        categories = {sorted_categories[-1]} if sorted_categories else set()
+        categories = {latest_categories[0]} if latest_categories else set()
```
squeaky-pl, line 115: This is now sorted in DESCending order so that's why -1 becomes 0.

squeaky-pl:
The affected code is already well covered by tests.

drewler (approved):

PR #971, opened 2024-11-07
Batch and optimize large deletions

squeaky-pl:
After https://github.com/closeio/sync-engine/pull/969 we got RAM and CPU usage under control.

The other problem with this code is that it can be painfully slow, especially with large deletions.

I know we have at least on escalation because of the slowness of this code where customer is using automation to move large amounts of emails between folders.

There are two problems here:
- For non conforming IMAP servers that can list one message under many uids we were deleting uids one by one and then updating a message, for such cases it's thousand of times much faster to delete all the uids in one go and then update message.
- We had N+1 queries when recalculating email categories

### DIFF
--- a/inbox/mailsync/backends/imap/common.py
+++ b/inbox/mailsync/backends/imap/common.py
@@ -12,11 +12,12 @@ accounts.
 
 """
 
+from collections import defaultdict
 from datetime import datetime
-from typing import List, Set
+from typing import List, Set, assert_never
 
 from sqlalchemy import bindparam, desc
-from sqlalchemy.orm import Query, Session
+from sqlalchemy.orm import Query, Session, joinedload
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.sql.expression import func
 
@@ -24,11 +25,14 @@ from inbox.contacts.processing import update_contacts_from_message
 from inbox.crispin import RawMessage
 from inbox.logging import get_logger
 from inbox.models import Account, ActionLog, Folder, Message, MessageCategory
-from inbox.models.backends.imap import ImapFolderInfo, ImapUid
+from inbox.models.account import CategoryType
+from inbox.models.backends.imap import ImapFolderInfo, ImapUid, LabelItem
 from inbox.models.category import Category
+from inbox.models.label import Label
 from inbox.models.session import session_scope
 from inbox.models.util import reconcile_message
 from inbox.sqlalchemy_ext.util import get_db_api_cursor_with_query
+from inbox.util.itert import chunk
 
 log = get_logger()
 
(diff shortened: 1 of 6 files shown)


inbox/mailsync/backends/imap/common.py
```diff
@@ -230,69 +253,109 @@ def remove_deleted_uids(account_id, folder_id, uids):
...
         return
-    deleted_uid_count = 0
-    for uid in uids:
-        # We do this one-uid-at-a-time because issuing many deletes within a
-        # single database transaction is problematic. But loading many
-        # objects into a session and then frequently calling commit() is also
-        # bad, because expiring objects and checking for revisions is O(number
-        # of objects in session), resulting in quadratic runtimes.
-        # Performance could perhaps be additionally improved by choosing a
-        # sane balance, e.g., operating on 10 or 100 uids or something at once.
```
squeaky-pl, line 241: This comment says it all, it was here from the very start. We no longer load as many objects to session as we limited that in previous PR.

inbox/mailsync/backends/imap/common.py
```diff
@@ -230,69 +253,109 @@ def remove_deleted_uids(account_id, folder_id, uids):
...
+    log.info(
+        "Deleted expunged UIDs",
+        uid_count=len(uids),
+        message_count=len(uids_by_message_id),
+    )
+
+
+def update_message_after_uid_deletion(
+    *, account_id: int, message_id: int, category_type: CategoryType
+) -> None:
```
squeaky-pl, line 311: This was just extracted to separate function, but git diff is useless here.

tests/imap/test_delete_handling.py
```diff
@@ -37,7 +37,7 @@ def test_messages_deleted_asynchronously(
         db.session,
     )
     assert "label" in [cat.display_name for cat in message.categories]
-    remove_deleted_uids(default_account.id, folder.id, [msg_uid])
+    remove_deleted_uids(default_account.id, folder.id, {msg_uid})
```
squeaky-pl, line 40: These are sets in practice everywhere in real code

inbox/models/backends/imap.py
```diff
@@ -209,11 +209,19 @@ def update_labels(self, new_labels: List[str]) -> None:
...
+
+        Args:
+            category_type: The category type for the account this uid belongs to.
+        """
+        if category_type == "label":
+            return {label.category for label in self.labels} | {self.folder.category}
+        elif category_type == "folder":
+            return {self.folder.category}
+        else:
+            assert_never(category_type)
```
squeaky-pl, line 224: Here we were unconditionally loading labels, even if the account does not support labels (only Gmail supports them), this was just stupid and I parametrized this on category_type to avoid this.

squeaky-pl: Additionally this would be a perfect place to use match and case but I need to update our lint tooling to support that syntax and there were too many things to change to make it work on this PR.

squeaky-pl:
These paths are already well tested.

drewler (approved):

PR #977, opened 2024-11-13
Recategorize messages

squeaky-pl:
Fixes #

Summary:

Test Plan:

Reviewers:


bin/recategorize-messages.py
```diff
@@ -0,0 +1,145 @@
...
+        changed_counter = 0
+        for message_id in message_ids:
+            with session_factory() as session:
+                message = session.query(Message).get(message_id)
+                old_categories = set(
+                    category.name for category in message.categories if category.name
+                )
+                update_message_metadata(
+                    session, message.account, message, message.is_draft
+                )
```
squeaky-pl, line 130: This is the function that was changed in a broken a way on the PRs that we reverted. So if we capture `categories` before, run the function after the reverts, and then read categories - and then are different we found a miscategorized email.

bin/recategorize-messages.py
```diff
@@ -0,0 +1,145 @@
...
+        f"Settings: {only_account_id=}, {only_inbox=}, {date_start=}, {date_end=}, {dry_run=}\n"
+    )
+
+    total_namespace_count = get_total_namespace_count(
+        only_account_id=only_account_id, only_types=set(only_types.split(","))
+    )
+    print(f"{total_namespace_count=}\n")
+
+    def session_factory():
+        return global_session_scope() if dry_run else session_scope(None)
```
squeaky-pl, line 107: This is sync-engine curiosity. sync-engine has two ways of managing transactions... `global_session_scope` is for reads and `session_scope` is mainly meant for writes. `session_scope` will automatically commit at the end. And it also automatically behind the scenes serializes changes to transaction log, and that transaction log is what CRM consumes.

bin/recategorize-messages.py
```diff
@@ -0,0 +1,145 @@
...
+    for account_id, namespace_id in account_id_to_namespace_id.items():
+        query = Query([Message.id]).filter(Message.namespace_id == namespace_id)
+
+        if only_inbox:
+            inbox_folder = ImapUid.folder.has(Folder._canonical_name == "INBOX")
+            query = query.filter(Message.imapuids.any(inbox_folder))
+        if date_start:
+            query = query.filter(Message.created_at >= date_start)
+        if date_end:
+            query = query.filter(Message.created_at < date_end)
```
neob91-close, line 74: Should we make it inclusive? Given a date range 2024-10-15 - 2024-10-16, I would imagine it's a 2 day interval.

neob91-close (approved):

bin/recategorize-messages.py
squeaky-pl: Yes, I understand your thinking. I tend to make it exclusive on one side because this way if you run two times [1, 2), [2, 3) you don't reprocess the intersection twice, but in this case we have a single interval.

[`3fc5c0a` (#977)](https://github.com/closeio/sync-engine/pull/977/commits/3fc5c0a0ca653ab5b386d094f9a99e48d5b962d7)

bin/recategorize-messages.py
```diff
@@ -0,0 +1,145 @@
...
+
+def yield_account_id_and_message_ids(
+    *,
+    only_account_id: int | None,
+    date_start: datetime.date | None,
+    date_end: datetime.date | None,
+    only_inbox: bool,
+    only_types: set[AccountType] = ALL_ACCOUNT_TYPES,
+) -> Iterable[int, list[int]]:
+    discriminators = {account_type + "account" for account_type in only_types}
```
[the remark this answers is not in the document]
squeaky-pl: [`98a3a22` (#977)](https://github.com/closeio/sync-engine/pull/977/commits/98a3a22352fb9e789a13ae196cda52eae963e6ea)

squeaky-pl:
@neob91-close I also changed `--only-account-id` to `--only-account-ids` in [`5eebb94` (#977)](https://github.com/closeio/sync-engine/pull/977/commits/5eebb949e4e869239d6ec7859a3d7c5af173640b) because I will target the customers that reported the problem first cluster by cluster before running it massively.

PR #984, opened 2024-12-10
Add pip compile

neob91-close:
I just wanted to introduce `pip-compile`, but had to switch to `ruff` as well, because we had some conflicting linting packages, and after fixing that, flake8 started reporting a whole lot of different errors.

We were going to introduce `ruff` anyway, I just wish we didn't have to do it all in one PR.

Definitely review commits separately.


.github/workflows/build-and-release.yaml
```diff
@@ -40,6 +40,34 @@ jobs:
       - name: Run flake8
         run: flake8
 
+  check-compiled-requirements:
```
neob91-close, line 40: Mostly copied from `closeio`

neob91-close:
Now I'm thinking maybe I should just swap out the current `*.txt` requirements files for the autogenerated ones as a separate PR.

squeaky-pl (approved):
The general structure looking good. Not reviewing machine generated changes.

tsx:
>  I just wish we didn't have to do it all in one PR.

What's stopping you from shipping ruff changes separately first?

Doing both ruff and dependency resolver change seems like an insane amount of risk to me.

neob91-close:
Ruff PR: https://github.com/closeio/sync-engine/pull/985

neob91-close:
Removed the ruff-related changes, and regenerated the requirements files.

PR #1001, opened 2024-12-13
Restore mypy

neob91-close:
Fixes #

Summary:

Test Plan:

Reviewers:


neob91-close:
In what way would you change the mypy settings?
Do you mean you'd make mypy more lenient so that we could add fewer type ignores?

Looks like we can maybe take the number of type ignores down from ~2000 to ~600 by manipulating the mypy config, but I don't think it's the better approach. I'd prefer stricter type-checking for all new code.

What drawbacks do you see in the number of type ignores?

tsx:
It's mostly the amount of extra noise.

The two main kinds of errors are:

- no-untyped-def - do you really need enforcement to keep adding annotations in new code? Ideally we could just annotate the all the existing stuff instead. It will take some effort for sure but I feel like it's very feasible.
- import-not-found - these are fixed by either updating the library or adding stubs. Sprinkling the codebase with type-ignores will just create more cleanup work when you do the update/stubs.

squeaky-pl (approved):
I see arguments for going both ways and I am not sure which way we want to go.

Are you going to be adding new code soon, like in next cycles?

I would at least run counts of each type-ignores. Create an issue which lists those in ascending order and put it to on your managers radar as a badly needed tech investment. Then each of type-ignore type can be tackled separately.

neob91-close:
FYI @maxpollack

PR #1015, opened 2025-02-17
Map "INBOX.Archiv" to "archive" in folder settings

mrhiggi-close:
Add support for the alternate folder name "INBOX.Archiv" to properly map it to the "archive" folder. This enhances compatibility with email providers using this naming convention.

We previously [added support for German "archiv"](https://github.com/closeio/sync-engine/pull/468) to the default folders. Custom providers use the `INBOX.Archive` format, and we now have a German custom email provider that uses `INBOX.Archiv`.

### DIFF
--- a/inbox/providers.py
+++ b/inbox/providers.py
@@ -443,6 +443,7 @@ providers: dict[str, dict[str, Any]] = {
         "auth": "password",
         "folder_map": {
             "INBOX.Archive": "archive",
+            "INBOX.Archiv": "archive",
             "INBOX.Drafts": "drafts",
             "INBOX.Junk Mail": "spam",
             "INBOX.Trash": "trash",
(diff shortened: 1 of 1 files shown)


inbox/providers.py
```diff
@@ -443,6 +443,7 @@ def provider_info(provider_name):  # type: ignore[no-untyped-def]  # noqa: ANN20
         "auth": "password",
         "folder_map": {
             "INBOX.Archive": "archive",
+            "INBOX.Archiv": "archive",
```
neob91-close, line 446: Not really sure if necessary, but would it make sense to add a brief comment that this is for some German providers?

neob91-close (approved):

squeaky-pl (approved):

squeaky-pl:
This might need manual intervention in the database, i.e. you should check in MySQL if the folder has correct role. If you don't want manual intervention you can suggest them to delete and readd their connected account.

inbox/providers.py
mrhiggi-close: Added a comment.

PR #1016, opened 2025-02-18
Handle additional Google API `insufficientPermissions` for calendar sync

mrhiggi-close:
Added support for handling the "insufficientPermissions" error in the Google API response. 

We saw a brief dip in overall error rate, but it looks like other errors were just being squashed by the rate limit. This is the next biggest offender.

### DIFF
--- a/inbox/events/google.py
+++ b/inbox/events/google.py
@@ -235,7 +235,11 @@ class GoogleEventsProvider(AbstractEventsProvider):
                         )
                         time.sleep(30 + random.randrange(0, 60))
                         continue
-                    elif reason in ["accessNotConfigured", "notACalendarUser"]:
+                    elif reason in [
+                        "accessNotConfigured",
+                        "notACalendarUser",
+                        "insufficientPermissions",
+                    ]:
                         self.log.warning(
                             f"API not enabled with reason {reason}; returning empty result"
                         )


mrhiggi-close:
Oh joy, CI Black doesn't like it. Time to fight with formatting.

neob91-close (approved):

neob91-close:
> We saw a brief dip in overall error rate

If this is in reference to the other PR, then it makes sense to link it here in case someone is reading this months or years from now.

squeaky-pl:
Googling reveals: 

> The insufficient permissions error is returned when you have not requested the scopes you need when you retrieved your access token. 

Does that mean that the upstream system that requests permission passes token without calendar scopes and we enabled calendar syncing downstream, because then the problem is on the other end and we shouldn't silence this.

How does this happen?
Can this be reproduced?

squeaky-pl:
Apparently you can retrieve scopes from a Google token using `https://www.googleapis.com/oauth2/v1/tokeninfo?access_token=xxxxxx`, can we retrieve a couple of tokens from sync-service on affected accounts and confirm we have calendar scopes?

squeaky-pl (commented):
Leaving comment to dismiss notification.

mrhiggi-close:
Closing until I can come back to this. I believe these are indeed tokens without calendar scope.

Questions to be answered:

* Are we still creating accounts with bad scope? Check the ages of problem accounts.
* What should we do to appropriately fail accounts lacking scope for a requested sync?
  * Presumably we shouldn't retry forever, but silently stopping isn't helpful. We should probably fail the account so the user can be notified? That'll impact email syncing, which may or may not be working for them at the moment.

PR #1043, opened 2025-06-17
Fix detection of invalid credentials

jpmelos:
## Main Event

This fixes detection of invalid credentials.

### Why was it failing?

Consider Script 1:

```python
from imapclient import IMAPClient

server = IMAPClient("imap.gmail.com", 993)
try:
    server.login("example@gmail.com", "qwertyasdfgh")
except IMAPClient.Error as exc:
    print(type(exc.args[0]), exc.args[0])
```

If you run Script 1, this is what you'll get:

```
<class 'str'> b'[AUTHENTICATIONFAILED] Invalid credentials (Failure)'
```

Note that `IMAPClient.Error().args[0]` is a stringified `bytes` object, it is enclosed in `b''`.

However, [the logic that we were using before to detect invalid credentials relied on the error message *starting with* a few patterns](https://github.com/closeio/sync-engine/blob/31d5558f8f9ebb279b08e963ea125ec13e8d1af2/inbox/auth/utils.py#L75), but it would never be the case because it *always* starts with `b'`.

### Why is the error message enclosed in `b''`?

I know, that's freaking maddening.

That's a bug in `IMAPClient`. If you look at [the implementation of `IMAPClient.login`, it calls `str(e)` when `e` is an exception coming from `imaplib`](https://github.com/mjs/imapclient/blob/46461f358f8410e110c09b63faf393c1b5b30090/imapclient/imapclient.py#L382), and it carries `bytes`.

And that's one very old bug! We're in version 2.2.0, which was released in January 2021, and the latest version of `imapclient` is 3.0.1, and [the bug is still there](https://github.com/mjs/imapclient/blob/391cc6d66d35c16bed11292ff00835646cd50ae5/imapclient/imapclient.py#L402)! The bug goes back to 2017, when it was shipped in [a commit that implemented "precise" exceptions](https://github.com/mjs/imapclient/commit/48807790fc9c6faeebf5fed2b2bfcaf87eac47dd).

## Side Cards

- Change the function name from `auth_is_invalid` to `is_error_message_invalid_auth` based in the best practice of naming functions that return booleans like yes-no questions. The new name is also more descriptive, indicating it expects an error message as input.
- There's no need to pass `exc` into that function, all it needs is the error message.
- A login failure is not an error by us, it is an error by the user of the software and we cannot fix it. It shouldn't be logged as an `error`, but instead as an `info` just so we have some paper trail in case we need to adjust some of the patterns we use to match authentication errors.
- A lot of the patterns we were matching before overlap. The set of things we match against can be simplified a lot.
  - Since this pattern is coming from the `IMAPClient.login` method, any errors we get from it are necessarily about logging in. If it contains words such as "invalid", "bad", "incorrect", etc, there's little doubt that this is about invalid credentials.
  - Have a test that tests against responses received from real servers.

### DIFF
--- a/tests/auth/test_utils.py
+++ b/tests/auth/test_utils.py
@@ -4,30 +4,27 @@ from inbox.auth.utils import is_error_message_invalid_auth
 
 
 @pytest.mark.parametrize(
-    ("error_message", "result"),
+    "error_message",
     [
-        # Server responses seen in the wild.
-        ("[authenticationfailed]", True),
-        ("incorrect username or password", True),
-        ("invalid login or password", True),
-        ("login login error password error", True),
-        ("[auth] authentication failed.", True),
-        ("invalid login credentials", True),
-        ("[ALERT] Please log in via your web browser", True),
-        ("LOGIN Authentication failed", True),
-        ("authentication failed", True),
-        ("[ALERT] Invalid credentials(Failure)", True),
-        ("Invalid email login", True),
-        ("failed: Re-Authentication Failure", True),
-        ("Invalid", True),
-        ("Login incorrect", True),
-        ("LOGIN GroupWise login failed", True),
-        ("LOGIN bad", True),
-        ("[AUTHORIZATIONFAILED]", True),
-        ("incorrect password", True),
-        # Strings not seen in the wild.
-        ("asdbadasd", False),  # Contains "bad", but not as a word.
+        "[authenticationfailed]",
+        "incorrect username or password",
+        "invalid login or password",
+        "login login error password error",
+        "[auth] authentication failed.",
+        "invalid login credentials",
+        "[ALERT] Please log in via your web browser",
+        "LOGIN Authentication failed",
+        "authentication failed",
+        "[ALERT] Invalid credentials(Failure)",
+        "Invalid email login",
+        "failed: Re-Authentication Failure",
+        "Invalid",
+        "Login incorrect",
+        "LOGIN GroupWise login failed",
+        "LOGIN bad",
+        "[AUTHORIZATIONFAILED]",
+        "incorrect password",
     ],
 )
-def test_auth_is_invalid(error_message, result):
-    assert is_error_message_invalid_auth(error_message) is result
+def test_auth_is_invalid(error_message):
+    assert is_error_message_invalid_auth(error_message) is True


inbox/auth/utils.py
```diff
@@ -37,42 +37,31 @@ def auth_requires_app_password(exc: IMAPClient.Error) -> bool:
...
-        "incorrect password",
+# IMAP doesn't have error semantics, so we have to match the error message
+# against a list of known responses to determine whether we couldn't log in
+# because the credentials are invalid. Sometimes a single pattern matches many
+# different responses from servers. For real error messages that have been
+# received from servers in the past, see the test suite.
+AUTH_INVALID_PATTERNS = [
+    pattern.lower()
+    for pattern in (
+        "fail",
```
neob91-close, line 48: Isn't this function about determining if the error message says that authentication failed due to invalid credentials?

I imagine an error message could contain "fail" or "bad" for other reasons?

inbox/auth/utils.py
```diff
@@ -37,42 +37,31 @@ def auth_requires_app_password(exc: IMAPClient.Error) -> bool:
...
 
-def auth_is_invalid(exc: IMAPClient.Error) -> bool:
-    if not exc.args:
-        return False
-    error_message = safe_decode(exc.args[0]).lower()
+def is_error_message_invalid_auth(error_message: str) -> bool:
+    normalized_error_message = error_message.lower()
     return any(
-        error_message.startswith(prefix) for prefix in AUTH_INVALID_PREFIXES
+        pattern in normalized_error_message
```
neob91-close, line 63: Should we wrap these patterns in regex `\b`? I imagine `bad` is such a short string that it could easily match part of a hexadecimal internal ID or something.

inbox/auth/utils.py
```diff
@@ -37,42 +37,31 @@ def auth_requires_app_password(exc: IMAPClient.Error) -> bool:
     )
 
 
-# IMAP doesn't really have error semantics, so we have to match the error
-# message against a list of known response strings to determine whether we
-# couldn't log in because the credentials are invalid, or because of some
```
neob91-close, line 42: I think losing `or because of some temporary server error` is a mistake, because it makes the purpose of this mechanism less clear.

inbox/auth/utils.py
jpmelos: Yea, that's a good idea. I'll do it.

inbox/auth/generic.py
```diff
@@ -74,16 +77,17 @@ def authenticate_imap_connection(  # type: ignore[no-untyped-def, override]
         try:
             conn.login(account.imap_username, account.imap_password)
         except IMAPClient.Error as exc:
-            if auth_is_invalid(exc):
-                log.error(
-                    "IMAP login failed", account_id=account.id, error=exc
-                )
-                raise ValidationError(exc)  # noqa: B904
+            if is_error_message_invalid_auth(exc.args[0]):
```
neob91-close, line 80: The original code in `auth_is_invalid` bailed out when `exc.args` was empty.
Should we do the same here?

jpmelos: It can't be empty. It can contain an empty string, which will still return `False`.

inbox/auth/utils.py
neob91-close: I read the entire PR description before reviewing. It's not about appeasing me, it's about making sure this logic doesn't return false positives. I can imagine an error like "Temporary server failure" that would be classified as an authentication error, while it shouldn't be.

inbox/auth/utils.py
neob91-close: I don't know, it just makes it more obvious what we're aiming for, when you add an "as opposed to" (btw, "or because" wasn't the best phrasing, and should probably have been "as opposed to").

I don't think it's that big of a deal, but I don't think it hurts to keep it. Feel free to ignore me here.

neob91-close (approved):

inbox/auth/utils.py
jpmelos: I think that's unnecessary. If I'm testing a string "to determine something", it is automatically implied that it is "as opposed to it not being that something".

inbox/auth/utils.py
jpmelos: That makes sense. I added some qualifiers for `fail`, and I also added a qualifier for `bad` since it could match the common "Bad Gateway" error. I don't think "invalid" or "incorrect" need qualifiers, but let me know if you disagree.

Done in https://github.com/closeio/sync-engine/pull/1043/commits/f8d9b3d97eaa11d85f6c29cd72a5b08eb9057ae2.

PR #1067, opened 2025-07-22
Add a health check API endpoint

neob91-close:
Depends on https://github.com/closeio/sync-engine/pull/1066

This adds a `/health` API endpoint on port 16384 that'll return HTTP 503 when:
- A certain amount of interruptible threads gets stuck
- A certain amount of time elapses between 240 and 360 minutes


jpmelos (approved):

squeaky-pl (approved):
Looks good. I would maybe deploy to one cluster first to see if it breaks something.

PR #1072, opened 2025-07-30
Fallback to uid when ical uid is none

mrhiggi-close:
Fixes https://github.com/closeio/sync-engine/issues/1071
Summary:
For some events, the `iCalUId` may be `None`. This causes a type error when synthesizing a cancellation event. It would also preclude events created after the cutoff date from being inserted into the events table.

We can fall back to the `uid`, accepting that this will allow duplicate event creation again, and cancellation events may have a different ID than their master event, if the `iCalUId` is not provided.

In my testing, cancellation events did include the `iCalUId`, but internet lore suggests it is possible that either of these scenarios may lead to missing ids:
* Hybrid exchange environments
* Cancellation events from the events delta api

### DIFF
--- a/inbox/events/microsoft/parse.py
+++ b/inbox/events/microsoft/parse.py
@@ -417,11 +417,13 @@ def synthesize_canceled_occurrence(
         + "-synthesizedCancellation-"
         + start_datetime.date().isoformat()
     )
-    cancellation_ical_uid = (
-        master_event["iCalUId"]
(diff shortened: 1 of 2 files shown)


squeaky-pl (approved):
Yes, this is probably possible.

I've seen people using old unsuported Microsoft Exchange servers or data that is migrated from such servers causing such weird edge cases.

PR #1075, opened 2025-07-31
555 upgrade requests to the latest version

EGrunzke:
- Updated requests to `2.32.4`
- Added `.cache` to `.gitignore` - pip puts a lot of files in there

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,6 @@ setup/
 
 # Pycharm
 .idea/
+
+# pip cache
+.cache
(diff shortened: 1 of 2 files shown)


EGrunzke:
> @EGrunzke asking just in case (because it's not mentioned in the PR description) - is this tested?

I ran the `test_data` script successfully. That sets up 6 emails accounts in the org, but I don't think any of them actually sync up. Is there a standard / preferred way to test syncing?

PR #1077, opened 2025-08-06
Use `passive_delete` to handle `LabelItem` cascade

mrhiggi-close:
When triggering a resync for gmail, we attempt to delete any unmatched imap UID entries during the `yield_chunk` iterations. If there is more than one chunk, then this query is still active in the session, and we get a `Commands out of sync; you can't run this command now` error.

Using `passive_deletes=True`, we can avoid lazy loading and deleting the associations in the ORM. We keep the `cascade="all, delete-orphan"` so that the ORM handles the deletion if we remove the label item from the association list.

This is the other approach I was considering in: https://github.com/closeio/sync-engine/pull/1063

### DIFF
--- a/tests/imap/test_folder_sync.py
+++ b/tests/imap/test_folder_sync.py
@@ -473,6 +473,8 @@ def test_gmail_handle_uidinvalid(
         .limit(5)
         .all()
     )
+
+    assert len(imap_uids) == 5
     for uid in imap_uids:
         uid.labels.add(label)
 


neob91-close:
I appreciate the detailed investigation! The `passive_deletes` solution seemed quite hacky to me, but I guess it's fine if there's no better way that works. I think I'd still prefer the batched solution (seems less brittle to me, and easier to understand). Are there any drawbacks you can think of?

That said, I don't want to block this too much longer, I'm sure you have better things to do than discuss different options on this one issue, so if you're adamant about the `passive_deletes`, I guess I'm okay with this.

tests/imap/test_folder_sync.py
```diff
@@ -428,3 +428,67 @@ def test_imap_message_deduplication(
...
+        default_account.email_address,
+        "gmail",
+        BoundedSemaphore(1),
+    )
+    folder_sync_engine.initial_sync()
+
+    # We need to add labels to our imap uids so that the cascade delete on
+    # LabelItems is required. This would trigger a deletion during the yield_chunk
+    # query if any ORM level cascades are triggered.
+    imap_uids = (
```
neob91-close, line 470: Nit: I think we should add a sanity check assertion that checks the number of `ImapUid` currently stored in the database is what we expect it to be. Currently it looks to me like we trust `initial_sync` to add stuff to the database, but I'd rather tests weren't so trusting.

neob91-close (approved):

neob91-close:
I think you may have run `black` with a different config than is run in CI.

I run linting this way (I know, hacky and inefficient):
```sh
ip-192-168-100-38:close-ui neob91$ cat ../sync-engine/lint
#!/bin/bash
docker compose run --rm app bash -c 'pip3 install -r requirements/requirements-lint.txt; ruff check --fix inbox; black inbox'
```

PR #1078, opened 2025-08-07
Fix ruff `G201` error

neob91-close:
Ruff G201 error docs: https://docs.astral.sh/ruff/rules/logging-exc-info/


inbox/models/event.py
```diff
@@ -523,11 +520,8 @@ def reconstruct(self) -> None:
         try:
             self.unwrap_rrule()
         except Exception as e:
-            log.error(  # noqa: G201
-                "Error parsing stored RRULE entry",
-                event_id=self.id,
-                error=e,
-                exc_info=True,
+            log.exception(
+                "Error parsing stored RRULE entry", event_id=self.id, error=e
```
jpmelos, line 524: Same.

inbox/sendmail/smtp/postel.py
```diff
@@ -161,9 +161,7 @@ def _connect(self, host, port):  # type: ignore[no-untyped-def]
             self.connection.connect(host, port)
         except OSError as e:
             # 'Connection refused', SSL errors for non-TLS connections, etc.
-            log.error(  # noqa: G201
-                "SMTP connection error", exc_info=True, server_error=e.strerror
-            )
+            log.exception("SMTP connection error", server_error=e.strerror)
```
jpmelos, line 164: Same.

inbox/sendmail/smtp/postel.py
```diff
@@ -364,9 +362,7 @@ def _send(self, recipients, msg):  # type: ignore[no-untyped-def]
                         )
             except smtplib.SMTPException as err:
                 last_error = err
-                self.log.error(  # noqa: G201
-                    "Error sending", error=err, exc_info=True
-                )
+                self.log.exception("Error sending", error=err)
```
jpmelos, line 365: Same.

squeaky-pl (commented):
I suppose we can drop exc_info=True as @jpmelos pointed out. I'm assuming this still works well with Sentry and reports stack trace correctly.

PR #1102, opened 2026-02-12
update pip-tools and wheel

saolsen:
Update pip-tools and explicit wheel version so it updates (which is a dependabot issue).


jpmelos (approved):
LGTM, but curious, why did you ask for a review from @closeio/messaging-scheduling-backend explicitly here? I feel like we don't have any more context than anybody else regarding this change.

PR #1109, opened 2026-03-12
`subqueryload` -> `selectinload`

jpmelos:
Replace all usage of `subqueryload` with `selectinload`. It appears that `selectinload` performs better and is more reliable than `subqueryload`. The results are the same in all possible scenarios and the interface is the same such that it's just a matter of replacing one name for the other.

- [ ] Deploy this to one cluster.
- [ ] Watch it for a while.

### DIFF
--- a/inbox/api/filtering.py
+++ b/inbox/api/filtering.py
@@ -8,7 +8,7 @@ from sqlalchemy import (  # type: ignore[import-untyped]
 )
 from sqlalchemy.orm import (  # type: ignore[import-untyped]
     contains_eager,
-    subqueryload,
+    selectinload,
 )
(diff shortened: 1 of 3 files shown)


wojcikstefan:
Just curious, why did you decide to close it @jpmelos? Was the premise laid out in the PR description misguided?

jpmelos:
I wrote this PR because I was firefighting a production issue. In the end we put out the fire with https://github.com/closeio/sync-engine/pull/1111. The premise laid out here is probably still good, but it's not a priority for me to merge it anymore and I'd rather not fix what is not broken because I may break it in the process.

PR #1110, opened 2026-03-12
Optimize DB query, move `order by` to client side

jpmelos:
The `order by` clause was making the planner choose not to use an index that is actually helpful for the query, so let's remove it and order client-side since the list of events returned here is very small (most often 0-1 elements).


thomasst (approved):

wojcikstefan:
Why close without merging after the PR has already been approved? Did we find a better way?

jpmelos:
We did this: https://github.com/closeio/sync-engine/pull/1111.

PR #1111, opened 2026-03-12
Load events for messages manually

jpmelos:
The `order by` clause is making the planner choose not to use an index that is actually helpful for the query to find events for messages, so let's remove find the events and insert them into the `Message` objects on the client side. Building the query manually opens up the possibility of using the `with_hint` API to request a `FORCE INDEX`.


drew-myers (approved):

mrhiggi-close (approved):
Building the query manually looks better to me than the statement rewriting.

PR #1127, opened 2026-04-13
Switch from black to ruff for formatting

marcospri:
Switch from black to ruff for autoformatting.

There's a black security alert, taking the opportunity to reformat with ruff and removed the black dependnecy.

Check pyproject.toml for config changes, it acounts for these (in the commit message):

```
warning: The following rules have been removed and ignoring them has no effect:
    - PT004
    - PT005

warning: `TCH005` has been remapped to `TC005`.
warning: `TRY302` has been remapped to `TRY203`.
The following rule may cause conflicts when used with the formatter: `ISC001`
incorrect-blank-line-before-class` (D203) and `no-blank-line-before-class` (D211) are incompatible. Ignoring `incorrect-blank-line-before-class`.
```


.git-blame-ignore-revs
```diff
@@ -1,3 +1,5 @@
 # .git-blame-ignore-revs
 # Blacken bin
-41eb2ccd36c119f377c87356c74a6559b6c69ca8
\ No newline at end of file
+41eb2ccd36c119f377c87356c74a6559b6c69ca8
+# Migrate auto-formatting from black to ruff
```
marcospri: This probably makes more sense after squash-merging the PR.

marcospri:
Not sure who to tag for a review, picking you @squeaky-pl based on the contribs. Feel free to ping someone else 👍 

Added a few differnt commits in case it's easier for review.

wojcikstefan:
I can take this one since I was doing some Sync Engine dependencies work recently and did the same transition in the Dialer codebase.

pyproject.toml
```diff
@@ -1,11 +1,10 @@
-[tool.black]
-skip-magic-trailing-comma = true
-target-version = ["py312"]
-line-length = 79
-
 [tool.ruff]
 target-version = "py312"
 line-length = 79
+exclude = [
```
wojcikstefan: [`exclude` ](https://docs.astral.sh/ruff/settings/#exclude) already contains `.venv` and `venv` by default. I suggest that we leave it default and if we need to exclude something extra, we use [`extend-exclude`](https://docs.astral.sh/ruff/settings/#extend-exclude) instead.

pyproject.toml
```diff
@@ -91,6 +85,13 @@ ignore = [
     "UP032",
     # Not ready for __future__ annotations
     "UP037",
+    # D203 and D211 are incompatible. Ignoring D203 
+    "D203",
+    # `import` should be at the top-level of a file
+    "PLC0415",
+    # The following rule may cause conflicts when used with the formatter
+    "ISC001"
```
wojcikstefan: I don't think this is true. [`ISC001`](https://docs.astral.sh/ruff/rules/single-line-implicit-string-concatenation/) only complains if, after a reformat, we're left with `"something"  "like this"` on a single line. That's indeed wrong and we should fix it as `"something like this"`.

wojcikstefan (approved):
Looks great! Happy to see this change.

pyproject.toml
marcospri: This is somewhat true as I copied it from the output of `ruff format` 😁:


> warning: The following rule may cause conflicts when used with the formatter: `ISC001`. To avoid unexpected behavior, we recommend disabling this rule, either by removing it from the `select` or `extend-select` configuration, or adding it to the `ignore` configuration.


But this doesn't happen in newer versions of ruff. I'll remove the ignore so we don't ignore this when we update, we can live with the warning for now.

I don't want to update ruff in this same PR, reformatting + new linting changes can get confusing.

PR #1142, opened 2026-05-14
Update lxml to 6.1.0

marcospri:
Fixes 

https://us-west-2.console.aws.amazon.com/inspector/v2/home?region=us-west-2#/findings?findingArn=arn%3Aaws%3Ainspector2%3Aus-west-2%3A372094135098%3Afinding%2Fb375ab262ae521c200e330a5a3db0cb9

I somehow was fixated that this was a ubuntu level dep and not a python one 🤦 .

Full changelog can be found in:

https://github.com/lxml/lxml/releases


marcospri:
From claude, with the change long in context:

```
The lxml usage in this repo is just two files, both calling ET.XML() and ET.tostring(). None of the changelog's breaking changes touch those APIs:

  - The XXE fix (6.1.0) only applies to iterparse() and ETCompatXMLParser — not used here. ET.XML() has used resolve_entities='internal' (safe) since lxml 5.0.
  - lxml.html.clean moved to a separate package (5.2.0) — not used.
  - MemDebug.dump/show() removed (6.0.0) — not used.
  - strip_cdata deprecated on HTMLParser (5.3.0) — the HTMLParser in inbox/util/html.py is Python stdlib's, not lxml's.
  - Python < 3.8 support removed (6.0.0) — not a concern.

  No compatibility changes are needed. The upgrade is clean as-is.
```

wojcikstefan (approved):

wojcikstefan:
We should observe Sentry closely after this deploys just in case, but I agree the changelog looks innocent given our usage.

PR #1160, opened 2026-07-09
Adjust tests for `strip_tags` after distro patch

jpmelos:
Context: https://github.com/closeio/sync-engine/pull/1159#issuecomment-4927891613


wojcikstefan (approved):

zbmott (approved):
Looks like a fun one!

PR #1164, opened 2026-09-01
Switch dependency managment to uv

marcospri:
Similar approach to:

- https://github.com/closeio/authalligator/pull/353
- https://github.com/closeio/socketshark/pull/354

refs: BEP-56


.github/workflows/build-and-release.yaml
```diff
@@ -18,59 +18,21 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
 
-      - name: Setup Python
```
marcospri, line 21: Bumping the GHA action version as in the authalligator counterpart.

.github/workflows/build-and-release.yaml
```diff
@@ -18,59 +18,21 @@ jobs:
...
-      - name: Install production requirements (for Mypy)
-        run: |
-          # Mypy needs production packages for typechecking
-          pip install --no-deps -r requirements/requirements-prod.txt
+        run: uv run ruff check --no-cache --output-format github .
 
       - name: Mypy
-        run: mypy
-
-  check-compiled-requirements:
```
marcospri, line 46: uv sync --locked should do this out of the box.

.github/dependabot.yml
```diff
@@ -3,10 +3,10 @@
 version: 2
 updates:
   # Enable version updates for python
-  - package-ecosystem: "pip"
-    directory: "/requirements"
+  - package-ecosystem: "uv"
```
marcospri, line 6: As with the other PRs taking the opportunity to reduce noise.

inbox/events/remote_sync.py
```diff
@@ -1,7 +1,7 @@
 from datetime import datetime, timedelta
 from typing import Any
 
-import more_itertools  # type: ignore[import-not-found]
```
marcospri, line 4: We now do `uv sync --locked` & `[{include-group="lint"}, {include-group="test"}]` so the test group is now installed with lint.

Before we just did 

- name: Install production requirements (for Mypy)
  run: pip install --no-deps -r requirements/requirements-prod.txt

so this was not present.

inbox/util/testutils.py
```diff
@@ -7,7 +7,7 @@
 
 import attr
 import dns
-import pytest  # type: ignore[import-not-found]
```
marcospri, line 10: Same as inbox/events/remote_sync.py

inbox/events/recurring.py
```diff
@@ -151,7 +151,7 @@ def get_start_times(  # type: ignore[no-untyped-def]  # noqa: ANN201
 
         if len(excl_dates) > 0:
             if not isinstance(rrules, rruleset):
-                rrules = rruleset().rrule(rrules)
```
marcospri, line 154: We don't pin types-python-dateutil (same as before) so this changed. IMO that's fine.

wojcikstefan: I think we should pin stuff for the sake of CI, no?

pyproject.toml
```diff
@@ -1,3 +1,105 @@
+[project]
+name = "sync-engine"
+version = "0.1.0"
+description = "Close's email sync engine."
+requires-python = "==3.12.*"
+dependencies = [
+    "alembic==1.7.5",
```
wojcikstefan, line 7: Are all of these direct dependencies? Or will we separately prune them so that the lockfile contains _all_ dependencies and this list only contains what we directly depend on?

pyproject.toml
```diff
@@ -1,3 +1,105 @@
...
+[dependency-groups]
+lint = [
+    "mypy==1.13.0",
+    "ruff==0.8.2",
+    "types-boto",
+    "types-python-dateutil",
+    "types-pytz",
+    "types-PyYAML",
+    "types-redis",
+    "types-requests",
```
wojcikstefan, line 71: Can we pin those? I don't think our CI should start failing just because something new got released.

wojcikstefan (approved):
<3

pyproject.toml
marcospri: Yeah I don't see a reason why these were not pinned in the first place. 

I'll do it in follow up to keep this PR changes to just pip-tools -> uv.

pyproject.toml
marcospri: Probably not, I'll do a pass cleaning it up, similar to: https://github.com/closeio/authalligator/pull/378

marcospri: There are probably some stale indirect dependencies there. I'll do a follow up PR with a cleanup.

wojcikstefan (approved):
🚢 🫡

PR #1186, opened 2026-09-17
Remove lint dependencies that are unused or non-direct

marcospri:
refs: BEP-56
follow up to: https://github.com/closeio/sync-engine/pull/1184

---
Removed:

types-boto we use boto3, this are for the older version 
types-request, also ships types

Pinned the version of the rest included types


wojcikstefan:
Seems like the mypy fails in CI

pyproject.toml
```diff
@@ -63,12 +63,10 @@ dependencies = [
 lint = [
     "mypy==1.13.0",
     "ruff==0.8.2",
-    "types-boto",
-    "types-python-dateutil",
-    "types-pytz",
-    "types-PyYAML",
-    "types-redis",
-    "types-requests",
+    "types-python-dateutil==2.9.0.20260807",
```
marcospri, line 66: Pin the types, it could lead to them getting out of sync and getting non-determinisict CI

marcospri:
> Seems like the mypy fails in CI

My bad, I think other repos default to create PRs in draft (or maybe that's based on my own behaviour?) 

Anyway ready for review now.

wojcikstefan (approved):

Issue #316, opened 2022-02-15
Get rid of care of Google/Apple contact code


squeaky-pl:
Code that reaches to Google/Apple services for contact data is defunct. Remove it.

Note that the code extract contact information from attached contact files (.vcf, .vcard) in emails is working correctly and we don't want to remove it.

wojcikstefan:
Why is that code defunct btw? Is it no longer allowed for integrations to fetch Google/Apple contacts via their APIs?

marcospri:
I arrived here via looking at this missing names:

https://github.com/closeio/sync-engine/blob/master/inbox/contacts/google.py#L184

This is very confusing, it looks like it should fail to load that module but maybe this is not used at all? I'd vote to remove it but ideally understanding what replaces its.

PR #1187, opened 2026-09-18
Remove unused & non-direct declared dependencies for prod

marcospri:
### Moved:
more-itertools, declared in test only, but its imported in prod, moved here.

### Removed:
asn1crypto - it was a dependency of cryptography, but not anymore
colorlog - not sure, not used
importlib_metadata  - no longer needed, we are in python 3.12, nothing uses them anyway
importlib_resources
zipp - was required by importlib_metadata
psutil - Not imported, not used

### Transitive dependencies:

request-file - Via tldextract
s3transfer - Via boto3
six - via a bunch of things


marcospri:
gdata is still declared and un-imported but technically is being used (?!) see https://github.com/closeio/sync-engine/issues/316

wojcikstefan (approved):

